PR #124, opened 2025-07-31
Feature/banner

softeng99:


libs/ui-react/src/lib/Components/Button/Button.tsx
```diff
@@ -21,7 +21,7 @@ const buttonVariants = cva(
       },
       size: {
         xs: 'px-12 py-8 body-2-semi-bold',
-        sm: 'px-16 py-8 body-2-semi-bold',
+        sm: 'px-16 py-12 body-2-semi-bold',
```
aammami-ledger, line 24: nice, I did not realise this was bot correct

aammami-ledger (approved):

PR #154, opened 2025-08-08
feat: checkbox component

softeng99:
- Added Checkbox component
- Added stories and documentation
- Added unit tests

### DIFF
--- a/libs/ui-react/src/lib/Components/Checkbox/Checkbox.stories.tsx
+++ b/libs/ui-react/src/lib/Components/Checkbox/Checkbox.stories.tsx
@@ -45,10 +45,9 @@ type Story = StoryObj<typeof meta>;
  * The base checkbox in its normal state.
  */
 export const Base: Story = {
-  args: {},
-  render: () => (
+  render: (args) => (
     <div className="flex h-80 w-80 items-center justify-center">
-      <Checkbox />
+      <Checkbox {...args} />
     </div>
   ),
 };


libs/ui-react/src/lib/Components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+};
+
+/**
+ * Example of checkbox usage in a form context.
+ */
+export const FormExample: Story = {
+  render: () => (
+    <form className="space-y-16 p-16">
+      <div className="space-y-12">
+        <h3 className="body-1">Subscribe to newsletters</h3>
```
aammami-ledger, line 94: Can you add `text-base` to all text section otherwise in dark mode it will be black on black

image

libs/ui-react/src/lib/Components/Checkbox/Checkbox.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+## Introduction
+
+Checkboxes are form controls that allow users to select one or multiple options from a list. They provide clear visual feedback about the selection state and support various interaction states including hover, focus, and disabled modes.
+
+> View in [Figma](https://www.figma.com/design/JxaLVMTWirCpU0rsbZ30k7/2.-Components-Library?node-id=90-1197&m=dev).
+
+## Anatomy
+
+<Canvas of={CheckboxStories.Base} />
```
aammami-ledger, line 17: this story has a small display issue 🤔 the hight of the story slitty moves between checked an unchecked states 

https://github.com/user-attachments/assets/28e82c2c-3afb-4908-943d-7c8de6ed73af

libs/ui-react/src/lib/Components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,101 @@
...
+  'h-20 w-20 shrink-0 rounded-xs transition-colors',
+  'focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-focus focus-visible:ring-offset-2',
+
+  // Unchecked states
+  'data-[state=unchecked]:border data-[state=unchecked]:border-muted data-[state=unchecked]:bg-base',
+  'data-[state=unchecked]:hover:bg-base-hover',
+  'data-[state=unchecked]:active:bg-base-pressed',
+  'data-[state=unchecked]:disabled:border-disabled data-[state=unchecked]:disabled:bg-base',
+
+  // Checked states
```
aammami-ledger, line 17: Can we drop these ? feels useless 
```suggestion
```

aammami-ledger:
> > Looking at the[ Shadcn checkbox ](https://ui.shadcn.com/docs/components/checkbox)I see they are making the whole line clickable (checkbox + label ), I think we would want that as well no ? to validate with @LaurineDevergne and @blucet-ledger
> 
> yeah, good point. currently the Checkbox component doesn't include the label, so it's up to the consumers to handle this, but maybe we should include it.

if it's up the consumers let maybe add an example on how they could do that  in storybook ?

softeng99:
> > > Looking at the[ Shadcn checkbox ](https://ui.shadcn.com/docs/components/checkbox)I see they are making the whole line clickable (checkbox + label ), I think we would want that as well no ? to validate with @LaurineDevergne and @blucet-ledger
> > 
> > 
> > yeah, good point. currently the Checkbox component doesn't include the label, so it's up to the consumers to handle this, but maybe we should include it.
> 
> if it's up the consumers let maybe add an example on how they could do that in storybook ? and I will ask Laurine to understand more what UX are er aiming for

I updated the form example to use labels associated with the checkboxes. please let me know if that's good enough or you had something else in mind.

aammami-ledger (approved):
LGTM! 
Just one note to fix before merging please: the controls are not working in storybook 

image

PR #159, opened 2025-08-19
feat: Link component

softeng99:
- Implemented Link component
- Fixed Button outline offset
- Removed focus ring from Button and CardButton (only use outline)

### DIFF
--- a/libs/ui-react/src/lib/Components/Link/Link.stories.tsx
+++ b/libs/ui-react/src/lib/Components/Link/Link.stories.tsx
@@ -113,8 +113,7 @@ export const External: Story = {
     appearance: 'underlined',
     children: 'External Link',
     isExternal: true,
-    href: 'https://example.com',
-    onClick,
+    href: 'https://ledger.com',
   },
   parameters: {
     docs: {
@@ -123,7 +122,7 @@ export const External: Story = {
 <Link
   appearance="underlined"
   isExternal
-  href="https://example.com"
+  href="https://ledger.com"
 >
   External Link
 </Link>
@@ -139,8 +138,7 @@ export const ExternalLinkWithIcon: Story = {
     children: 'External Link with Icon',
     icon: Information,
     isExternal: true,
-    href: 'https://example.com',
-    onClick,
+    href: 'https://ledger.com',
   },
   parameters: {
     docs: {
@@ -150,7 +148,7 @@ export const ExternalLinkWithIcon: Story = {
     appearance="underlined"
     icon={Information}
     isExternal
-    href="https://example.com"
+    href="https://ledger.com"
   >
     External Link with Icon
   </Link>


libs/ui-react/src/lib/Components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,317 @@
...
+          Medium Accent
+        </Link>
+      </div>
+    </>
+  ),
+};
+
+export const ResponsiveLayout: Story = {
+  render: () => (
+    <div className="flex w-256 flex-col gap-8 rounded-sm bg-canvas p-8">
```
aammami-ledger, line 242: can you please change the bg color of the container , it's white on white
```suggestion
    <div className="flex w-256 flex-col gap-8 rounded-sm bg-muted p-8">
```
image

libs/ui-react/src/lib/Components/Link/Link.tsx
```diff
@@ -0,0 +1,146 @@
...
+        ref={ref}
+        className={cn(
+          className,
+          linkVariants({
+            appearance,
+            size,
+          }),
+        )}
+        target={isExternal && !asChild ? '_blank' : undefined}
+        rel={isExternal && !asChild ? 'noopener noreferrer' : undefined}
```
aammami-ledger, line 120: if they want to add their own `target` and `rel` we should use them not override them no ?

target={isExternal && !asChild ? '_blank' : props.target}
rel={isExternal && !asChild ? 'noopener noreferrer' : props.rel}

softeng99: it's not very clear currently, but if they provide `target` and `rel` props it will overwrite the ones we set in the component. they are just the defaults. does that make sense? or should we enforce `_blank` and `noopener noreferrer` for external links?

aammami-ledger: but it will be confusing if they set a target and it won't work :/ am thinking in case someone wants an external link that opens in the same tab for instance

aammami-ledger: hmm interesting the code says differently or my mind has still not reset from holidays, So if someone wants an external link that opens in the same tab they can still do that ? by adding `target="_self"` ?

libs/ui-react/src/lib/Components/Link/Link.tsx
```diff
@@ -0,0 +1,146 @@
...
+                size={calculatedIconSize}
+                className="flex-shrink-0"
+              />
+            )}
+            <span className="min-w-0 truncate">{children}</span>
+            {isExternal && (
+              <ExternalLink
+                size={calculatedIconSize}
+                className="flex-shrink-0"
+              />
```
aammami-ledger, line 138:
```suggestion
              <>
				<ExternalLink
				size={calculatedIconSize}
				className="flex-shrink-0"
				aria-hidden="true"
				/>
				{/* Hidden text for screen readers */}
				<span className="sr-only">(opens in a new tab)</span>
              </>
```

would be nice to add  opens in a new tab to screen readers as the icon is not informative

PR #161, opened 2025-08-25
Feature/list item

softeng99:
- Implemented ListItem component
- Added implementation details, usage examples and stories
- Update Tag component so internal styles have precedence over the `className` prop


libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { IconSize } from '../Icon/Icon';
+import { Tag, type TagProps } from '../Tag/Tag';
+
+export interface ListItemProps
+  extends Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'children'> {
+  title: string;
+  description?: string;
+  descriptionTagProps?: TagProps;
+  leadingIcon?: React.ComponentType<{ size?: IconSize; className?: string }>;
+  trailingContent?: React.ReactNode;
```
aammami-ledger, line 12: while working on the input I wanted to add description to the props so it helps the user understand what is it about when you hover over it, I think we should do this often for props that are atypical like this  `trailingContent`
```suggestion
  /**
   * Custom content to render on the right side of the List-item.
   * @example trailingContent={<Icon />}
   */
  trailingContent?: React.ReactNode;
```
image

libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @component
+ * @param {string} title - The main title of the list item.
+ * @param {string} [description] - Optional descriptive text below the title.
+ * @param {TagProps} [descriptionTagProps] - Optional props for a tag displayed next to the description.
+ * @param {React.ComponentType<{ size?: IconSize; className?: string }>} [leadingIcon] - Optional icon to display on the left.
+ *   The icon styles are defined by the list item. Please do not override them.
+ * @param {React.ReactNode} [trailingContent] - Optional trailing content to display on the right side of the list item.
+ * @param {string} [className] - Additional custom CSS classes to apply. Do not use this prop to modify the component's core appearance.
+ * @param {React.ButtonHTMLAttributes<HTMLButtonElement>} [...] - All standard button props (e.g., `disabled`, `onClick`, etc.).
```
aammami-ledger, line 30: and now we can remove these and reduce the content of the JSdocs, WDYT ?

libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+                  className={cn(
+                    'truncate text-muted body-3',
+                    props.disabled && 'text-disabled',
+                  )}
+                >
+                  {description}
+                </div>
+                {descriptionTagProps && (
+                  <Tag
+                    {...descriptionTagProps}
```
aammami-ledger, line 109: I am not a big fan of this `descriptionTagProps`  prop honestly .. I thought the tag will have to have the same appearance as the list item, which is the `base` tag

aammami-ledger: maybe this is an other sign to make it a react element

libs/ui-react/src/lib/Components/ListItem/ListItem.stories.tsx
```diff
@@ -0,0 +1,296 @@
...
+      description: 'Optional trailing content to display on the right side',
+    },
+    onClick: { action: 'clicked' },
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof ListItem>;
+
+export const OnlyTitle: Story = {
```
aammami-ledger, line 65: I don't think we ever want the customers to use this with only a title let's remove these states and keep the minimal one we see in figma which is icon title and description

libs/ui-react/src/lib/Components/ListItem/ListItem.stories.tsx
```diff
@@ -0,0 +1,296 @@
...
+  },
+};
+
+export const WithLeadingIcon: Story = {
+  args: {
+    title: 'Item with Leading Icon',
+    leadingIcon: Settings,
+    className: 'max-w-256',
+  },
+};
```
aammami-ledger, line 95: same, let's remove these

aammami-ledger (commented):
Let's add a control for the `trailingContent` and description for the `onclick`

image

libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+        className={cn(
+          'flex h-64 w-full cursor-pointer items-center gap-16 rounded-md bg-base-transparent px-8 py-12 text-base',
+          'hover:bg-base-transparent-hover focus-visible:outline-2 focus-visible:outline-focus active:bg-base-transparent-pressed',
+          'disabled:cursor-default disabled:bg-base-transparent disabled:text-disabled',
+          className,
+        )}
+        {...buttonProps}
+      >
+        <div className="flex min-w-0 flex-1 items-center gap-12">
+          {LeadingIcon && <LeadingIcon size={48} className="shrink-0" />}
```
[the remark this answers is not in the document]
softeng99: yeah, makes sense. not sure about the naming in this case 🤔  maybe `leadingContent`?

aammami-ledger (approved):
Approved! 

I think it will be cool to add an exemple on how to use this with a Spot not only icons, would be useful as that's what we have on the send designs 

image

we can do that on a follow up once spot is merged!

PR #165, opened 2025-08-27
feat: add Input component with implementation, documentation, and tests

aammami-ledger:
## Input Component 

- Introduced a new baseInput: Internal usage only
- Introduced a new Input component on top of baseInput featuring a floating label, clear button, and error handling.
- Added comprehensive documentation and usage examples in Storybook.
- Implemented unit tests to ensure functionality and accessibility compliance.

### DIFF
--- a/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
+++ b/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
@@ -80,7 +80,11 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
         ? true
         : undefined;
 
-    const hasContent = !!props.value && props.value.toString().length > 0;
+    const isControlled = props.value !== undefined;
+    const hasContent = isControlled
+      ? !!props.value && props.value.toString().length > 0
+      : !!inputRef.current?.value;
+
     const showClearButton = hasContent && !disabled && onClear;
 
     const errorId = `${inputId}-error`;
--- a/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
+++ b/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
@@ -159,7 +159,16 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
             /** TODO: extract to a new component IconButton */
             <button
               type="button"
-              onClick={onClear}
+              onClick={() => {
+                if (isControlled) {
+                  onClear?.();
+                } else if (inputRef.current) {
+                  inputRef.current.value = '';
+                  const event = new Event('input', { bubbles: true });
+                  inputRef.current.dispatchEvent(event);
+                  onClear?.();
+                }
+              }}
               className="mr-16 cursor-pointer rounded-full"
               aria-label="Clear input"
               data-side="right"
--- a/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
+++ b/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { cn } from '@ldls/utils-shared';
 import { DeleteCircleFill } from '../../Symbols';
 
-const baseRootStyles = cn(
+const baseContainerStyles = cn(
   'group relative flex h-48 w-full items-center gap-8 px-16 rounded-md bg-muted transition-colors',
   'hover:bg-muted-hover focus-within:ring-2 focus-within:ring-active',
   'has-[:disabled]:pointer-events-none has-[:disabled]:cursor-not-allowed has-[:disabled]:bg-disabled has-[:disabled]:text-disabled',
@@ -119,7 +119,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
         // Always call the original onChange if provided
         onChangeProp?.(e);
       },
-      [isControlled, this, onChangeProp],
+      [isControlled, onChangeProp],
     );
 
     const hasContent = isControlled
@@ -170,7 +170,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
     return (
       <div>
         <div
-          className={cn(containerClassName, baseRootStyles)}
+          className={cn(containerClassName, baseContainerStyles)}
           onPointerDown={(event: React.PointerEvent<HTMLDivElement>) => {
             const target = event.target as Element;
             if (target.closest('input, button, a')) return;
--- a/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
+++ b/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
@@ -70,7 +70,8 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
   ) => {
     const inputRef = React.useRef<HTMLInputElement>(null);
 
-    const inputId = id || `input-${Math.random().toString(36).slice(2, 11)}`;
+    const reactId = React.useId();
+    const inputId = id || `input-${reactId}`;
 
     // Handle aria-invalid properly - use provided value or derive from errorMessage
     const ariaInvalid = props['aria-invalid']
--- a/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
+++ b/libs/ui-react/src/lib/Components/Input/BaseInput.tsx
@@ -3,7 +3,7 @@ import { cn } from '@ldls/utils-shared';
 import { DeleteCircleFill } from '../../Symbols';
 
 const baseRootStyles = cn(
-  'group relative flex h-48 w-full items-center rounded-md bg-muted transition-colors',
+  'group relative flex h-48 w-full items-center gap-8 px-16 rounded-md bg-muted transition-colors',
   'hover:bg-muted-hover focus-within:ring-2 focus-within:ring-active',
   'has-[:disabled]:pointer-events-none has-[:disabled]:cursor-not-allowed has-[:disabled]:bg-disabled has-[:disabled]:text-disabled',
   'has-[:invalid]:ring-1 has-[:invalid]:ring-error has-[:invalid]:border-error',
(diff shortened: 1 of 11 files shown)


libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,188 @@
...
+              onClick={onClear}
+              className="mr-16 cursor-pointer rounded-full"
+              aria-label="Clear input"
+              data-side="right"
+            >
+              <DeleteCircleFill
+                size={20}
+                className={cn('text-muted', ariaInvalid && 'text-error')}
+              />
+            </button>
```
aammami-ledger, line 164: not sure what to do about this clear button, we would probably need a component that covers these kind of situations

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,188 @@
...
+            id={inputId}
+            disabled={disabled}
+            placeholder=" "
+            aria-invalid={ariaInvalid}
+            aria-describedby={errorMessage ? errorId : undefined}
+            className={cn(
+              baseInputStyles,
+              label && 'pt-16',
+              rightElement && 'pr-8',
+              leftElement && 'pl-8',
```
softeng99, line 141: can we handle all the spacing on the container instead, using padding and gap?

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,188 @@
...
+            <label htmlFor={inputId} className={baseLabelStyles}>
+              {label}
+            </label>
+          )}
+
+          {showClearButton && (
+            <button
+              type="button"
+              onClick={onClear}
+              className="mr-16 cursor-pointer rounded-full"
```
softeng99, line 156: similarly here for the margin

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
softeng99: yeah, I agree. maybe an `IconButton`?

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const inputId = id || `input-${Math.random().toString(36).slice(2, 11)}`;
+
+    // Handle aria-invalid properly - use provided value or derive from errorMessage
+    const ariaInvalid = props['aria-invalid']
+      ? props['aria-invalid']
+      : errorMessage
+        ? true
+        : undefined;
+
+    const hasContent = !!props.value && props.value.toString().length > 0;
```
softeng99, line 82: should we handle uncontrolled inputs as well? something like this:

```
const isControlled = props.value !== undefined;
const hasContent = () => isControlled
  ? !!props.value && props.value.toString().length > 0
  : inputRef.current.value;
```

although in this case `onClear` wouldn't work unless we provide a prop like `withClearButton` or something instead of just relying on `onClear`. what do you think?

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,188 @@
...
+      return (node: T) => {
+        refs.forEach((ref) => {
+          if (!ref) return;
+          if (typeof ref === 'function') {
+            ref(node);
+          } else {
+            (ref as React.MutableRefObject<T | null>).current = node;
+          }
+        });
+      };
```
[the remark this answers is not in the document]
aammami-ledger: I'd rather do this in a follow up PR, as this would be the first time we use ui-core, and just moving the code there and building the packages did not work out of the box so I need to make some changes .. will rather so that as a follow up

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
aammami-ledger: gap alone won't fix, the thing is I need to update the main padding of the input (not the spacing between the input elements )

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
aammami-ledger: added a TODO for this, I do not want to overload this one

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
aammami-ledger: great comment! I have totally missed this one, what about [this](https://github.com/LedgerHQ/ldls/pull/165/commits/535cd94cbaf10b02b643adaf89dde33d64a5e4d3) ?

softeng99: yeah this should work, but for uncontrolled inputs users would have to provide an "empty" function (`onClear={() => {}}` in order to render the clear button. it feels a bit weird to me

aammami-ledger: it's either that or we can make a strategic change:
Always show the clear button when there’s content: controlled inputs rely on `onClear` for the action, while uncontrolled inputs clear themselves automatically.

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,238 @@
...
+    const handleClear = () => {
+      if (!inputRef.current) return;
+
+      // Default clear behavior
+      if (isControlled && props.onChange) {
+        // Controlled → trigger onChange with empty value
+        props.onChange({
+          target: { value: '' },
+          currentTarget: { value: '' },
+        } as React.ChangeEvent<HTMLInputElement>);
```
aammami-ledger, line 122: To discuss

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,238 @@
...
+      } else {
+        // Uncontrolled → update state and DOM
+        inputRef.current.value = '';
+        setUncontrolledValue('');
+        const event = new Event('input', { bubbles: true });
+        inputRef.current.dispatchEvent(event);
+      }
+      inputRef.current.focus();
+      // Call onClear if provided
+      onClear?.();
```
aammami-ledger, line 132: should we extend the `onClear` or override it if a `onClear` function is provided ?

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
aammami-ledger: done

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
aammami-ledger: done

aammami-ledger: done

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,238 @@
...
+      ? props['aria-invalid']
+      : errorMessage
+        ? true
+        : undefined;
+
+    const isControlled = props.value !== undefined;
+
+    const [uncontrolledValue, setUncontrolledValue] = React.useState(
+      props.defaultValue?.toString() || '',
+    );
```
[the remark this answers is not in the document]
aammami-ledger: When `defaultValue` is defined `inputRef.current.value.length > 0` returns false --> close icon do not appear

I have added `uncontrolledValue` and initialised it with `defaultValue` so if we have no value in our ref but we do have a default one the close ion will appear

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+          {label && (
+            <label htmlFor={inputId} className={baseLabelStyles}>
+              {label}
+            </label>
+          )}
+
+          {showClearButton && (
+            /** TODO: extract to a new component IconButton */
+            <button
```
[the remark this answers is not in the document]
aammami-ledger: should be handled inside the new IconComponent you are working on, I've added a todo to replace it with it once it's done

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,238 @@
...
+        // Uncontrolled → update state and DOM
+        inputRef.current.value = '';
+        setUncontrolledValue('');
+        const event = new Event('input', { bubbles: true });
+        inputRef.current.dispatchEvent(event);
+      }
+      inputRef.current.focus();
+      // Call onClear if provided
+      onClear?.();
+    };
```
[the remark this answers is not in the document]
aammami-ledger: as _valueTracker is internal, undocumented, and not stable. React could change or remove it in future versions :eyes:

I opted for a diffrent approach, WDYT ?

```
      // This is a robust way to programmatically trigger an onChange for controlled components.
      // It simulates a user action more closely.
      const valueSetter = Object.getOwnPropertyDescriptor(window.HTMLInputElement.prototype, 'value')?.set;
      valueSetter?.call(input, '');

      const event = new Event('input', { bubbles: true });
      input.dispatchEvent(event);
```

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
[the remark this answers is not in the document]
aammami-ledger: ah yeah this will make hasContent always `true` 

https://github.com/user-attachments/assets/d9817c33-0a2a-4650-baba-6dceea84cd4e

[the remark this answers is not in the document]
aammami-ledger: The issue with this is that hasContent will remain `true` 
- we cleared the input → DOM value became empty 
- But React never re-rendered to recalculate hasContent
- Clear button stayed visible despite empty input we are setting 

https://github.com/user-attachments/assets/e36a2f80-03ba-4dbb-9dab-f870bbc6919b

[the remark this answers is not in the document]
aammami-ledger: sure good idea

libs/ui-react/src/lib/Components/Input/BaseInput.tsx
```diff
@@ -0,0 +1,249 @@
...
+          className={cn(containerClassName, baseRootStyles)}
+          onPointerDown={(event: React.PointerEvent<HTMLDivElement>) => {
+            const target = event.target as Element;
+            if (target.closest('input, button, a')) return;
+
+            const input = inputRef.current;
+            if (!input) return;
+
+            const isSuffix = target.closest('[data-side="right"]');
+            const cursorPosition = isSuffix ? input.value.length : 0;
```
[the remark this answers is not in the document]
aammami-ledger: very smart 🧠  DONE!

softeng99 (approved):
Nice work! This was a tricky one! 🎉

PR #168, opened 2025-08-29
feat: Spot component

softeng99:
- Implemented Spot component

### DIFF
--- a/libs/ui-react/src/lib/Components/Spot/Spot.tsx
+++ b/libs/ui-react/src/lib/Components/Spot/Spot.tsx
@@ -1,0 +1,152 @@
+import { cva } from 'class-variance-authority';
+import { IconSize } from '../Icon/Icon';
+import { HTMLAttributes, useMemo } from 'react';
+import {
+  Bluetooth,
+  CheckmarkCircleFill,
+  DeleteCircleFill,
+  InformationFill,
+  Spinner,
+  WarningFill,
+} from '../../Symbols';
+import { cn } from '@ldls/utils-shared';
+
+const spotVariants = cva(
+  'flex size-48 items-center justify-center rounded-full bg-muted-transparent',
+  {
+    variants: {
+      appearance: {
+        icon: 'text-base',
+        bluetooth: 'text-[#0082FC]',
+        check: 'text-success',
+        error: 'text-error',
+        warning: 'text-warning',
+        info: 'text-muted',
+        loader: '',
+        number: 'text-base heading-4',
+      },
+      disabled: {
+        true: 'text-disabled',
+      },
+    },
+  },
+);
+
+export type SpotAppearance =
+  | 'icon'
+  | 'bluetooth'
+  | 'check'
+  | 'error'
+  | 'warning'
+  | 'info'
+  | 'loader'
+  | 'number';
+
+type DiscriminatedSpotProps =
+  | {
+      appearance: 'icon';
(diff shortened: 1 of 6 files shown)


libs/ui-react/src/lib/Components/Spot/Spot.tsx
```diff
@@ -0,0 +1,152 @@
...
+  | 'number';
+
+type DiscriminatedSpotProps =
+  | {
+      appearance: 'icon';
+      icon: React.ComponentType<{ size?: IconSize; className?: string }>;
+    }
+  | {
+      appearance: 'number';
+      number: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;
```
aammami-ledger, line 52: NO two-digit numbers? Feels a bit strict for me

libs/ui-react/src/lib/Components/Spot/Spot.tsx
```diff
@@ -0,0 +1,152 @@
...
+    }
+  | {
+      appearance: Exclude<SpotAppearance, 'icon' | 'number'>;
+    };
+
+export type SpotProps = DiscriminatedSpotProps &
+  HTMLAttributes<HTMLDivElement> & {
+    disabled?: boolean;
+  };
```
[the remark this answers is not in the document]
aammami-ledger: and we have shorter description for main component functionality ;)

libs/ui-react/src/lib/Components/Spot/Spot.tsx
```diff
@@ -0,0 +1,152 @@
...
+        return <WarningFill size={20} />;
+      case 'info':
+        return <InformationFill size={20} />;
+      case 'loader':
+        return <Spinner size={20} />;
+    }
+  }, [props]);
+
+  return (
+    <div className={cn(className, spotVariants({ appearance, disabled }))}>
```
aammami-ledger, line 148: the component accepts `HTMLDivElement` but you are not propagating them is it intended ? I think we need to would be useful to be able to pass `id`, `onClick`, `data-testid` ect

libs/ui-react/src/lib/Components/Spot/Spot.tsx
softeng99: I asked Laurine about this because larger numbers wouldn't look very good and she confirmed it's only single digit numbers, at least for now.

libs/ui-react/src/lib/Components/Spot/Spot.tsx
```diff
@@ -0,0 +1,190 @@
...
+ * **For other appearances:** No additional props required. Each renders a predefined icon:
+ * - `'bluetooth'` - Bluetooth icon in active state
+ * - `'check'` - Check mark icon (success state)
+ * - `'error'` - Delete/error icon (error state)
+ * - `'warning'` - Warning icon (warning state)
+ * - `'info'` - Information icon
+ * - `'loader'` - Loading spinner icon
+ *
+ * @param {string} [className] - Additional custom CSS classes to apply. Do not use this prop to modify the component's core appearance - use the `appearance` prop instead.
+ * @param {HTMLAttributes<HTMLDivElement>} [...] - All standard div props
```
aammami-ledger, line 132: Let's remove these as we do added description above 

```suggestion
```

aammami-ledger (approved):

PR #175, opened 2025-09-02
refactor: update Banner and CardButton components to remove `isFull` 

aammami-ledger:
- Removed `isFull` prop from Banner and CardButton components, as they now naturally fill their parent container width.
- Updated documentation to clarify that container elements control the width of the components.
- Adjusted examples in implementation and stories to reflect the new width behavior.

### DIFF
--- a/libs/ui-react/src/lib/Components/CardButton/CardButton.stories.tsx
+++ b/libs/ui-react/src/lib/Components/CardButton/CardButton.stories.tsx
@@ -180,7 +180,7 @@ export const NaturalWidth: Story = {
 <CardButton
   appearance="base"
   title="Natural Width Example"
-  description="This card button demonstrates natural width behavior"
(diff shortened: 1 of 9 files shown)


libs/ui-react/src/lib/Components/CardButton/CardButton.stories.tsx
```diff
@@ -159,6 +159,36 @@ export const FullFeatures: Story = {
...
+    </div>
+  ),
+  parameters: {
+    docs: {
+      source: {
+        code: `
+<CardButton
+  appearance="base"
+  title="Natural Width Example"
+  description="This card button demonstrates natural width behavior"
```
aammami-ledger, line 183:
```suggestion
  description="This card button demonstrates how it flows naturally to fill the full width of its parent container without any max-width constraints"
```

PR #180, opened 2025-09-03
feat: add AddressFieldInput component

aammami-ledger:


libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+>(({ prefix, suffix, onQrCodeClick, ...props }, ref) => {
+  // Use custom prefix if provided, otherwise default "To:" prefix
+  const effectivePrefix = prefix || (
+    <span
+      className="text-nowrap text-base group-has-[:disabled]:text-disabled"
+      aria-hidden="true"
+    >
+      To:
+    </span>
+  );
```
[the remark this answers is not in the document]
softeng99: I think string should be fine, at least until someone screams at us

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { BaseInput, type BaseInputProps } from '../BaseInput';
+import { IconButton } from '../IconButton/IconButton';
+import { QrCode as QrCodeIcon } from '../../Symbols';
+
+export interface AddressFieldInputProps
+  extends Omit<BaseInputProps, 'prefix' | 'label'> {
+  /**
+   * Custom suffix element to show instead of the QR code icon.
+   * When provided, the QR code icon will not be shown.
+   */
```
softeng99, line 11: should we mention the clear button here too? so the default suffix is QR code (if the input is empty) or clear button (if the input is not empty), and both can be overwritten by providing a suffix explicitly

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+  suffix?: React.ReactNode;
+  /**
+   * Custom prefix element to show instead of the "To:" prefix.
+   * When provided, the "To:" prefix will not be shown.
```
softeng99, line 15: maybe this prop should default to `To:` (instead of doing something like `{prefix || 'To:'}`

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * ## Clear Button Behavior
+ * - Shows automatically when input has content and is not disabled
+ * - Works with both controlled and uncontrolled inputs using native value setter
+ * - Can be hidden with `hideClearButton={true}`
+ * - Extended behavior via optional `onClear` prop
+ *
+ * ## Layout & Spacing
+ * Uses container-based spacing (px-16 padding + gap-8) for consistent element positioning.
+ * Prefix (default "To:") is visible on the left, QR code icon appears when empty, clear button takes precedence when input has content.
```
softeng99, line 50: this is probably not needed

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+    <IconButton
+      iconType="stroked"
+      onClick={onQrCodeClick}
+      aria-label="Scan QR code"
+      disabled={!onQrCodeClick}
+      className="group-has-[:disabled]:text-disabled"
+    >
+      <QrCodeIcon size={20} />
+    </IconButton>
+  );
```
softeng99, line 108: should we only render the QR code if the input is empty? currently if `hideClearButton` is `true` the QR code will be rendered even if the input has content

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+ * />
+ */
+export const AddressFieldInput = React.forwardRef<
+  HTMLInputElement,
+  AddressFieldInputProps
+>(({ prefix, suffix, onQrCodeClick, ...props }, ref) => {
+  // Use custom prefix if provided, otherwise default "To:" prefix
+  const effectivePrefix = prefix || (
+    <span
+      className="text-nowrap text-base group-has-[:disabled]:text-disabled"
```
softeng99, line 89: I think this needs `body-1`

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Whether to hide the clear button that appears when input has content.
+   * @default false
+   */
+  hideClearButton?: boolean;
+  /**
+   * Callback fired when the QR code scanner icon is clicked.
+   * Only available when the input is empty and no custom suffix is provided.
+   */
+  onQrCodeClick?: () => void;
```
softeng99, line 27: should we only render the QR code if `onQrCodeClick` exists? otherwise we might need a `hideQrCode` prop

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
aammami-ledger: yeah good point I have thought about this and the way I see it is that if for some reason we decided to hide the closebutton it's because we want a permenant suffix, otherwise why hide it ? 

so given that logic I made the suffix permanent, but I do see your point you know it's hard at this stage to predict the usage of this

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.tsx
aammami-ledger: yeah fair point, made it only render when `onQrCodeClick` is passed

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.implementation.mdx
```diff
@@ -0,0 +1,267 @@
...
+<form>
+  <AddressFieldInput placeholder="Enter address or ENS" />
+</form>
+```
+
+### Address Format Support
+
+- **Ethereum addresses**: 42-character hexadecimal strings starting with '0x'
+- **ENS names**: Human-readable names ending with '.eth'
+- **Validation**: Implement real-time validation for better UX
```
softeng99, line 148: this might be confusing because we're not offering support for the address format, it's up to the consumers to validate the input. also this might be used for any chain, not just Ethereum.

libs/ui-react/src/lib/Components/AddressFieldInput/AddressFieldInput.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## Introduction
+
+AddressFieldInput is designed specifically where users need to enter wallet addresses or ENS names. It combines the reliability of BaseInput with domain-specific UX enhancements:
+
+- **Customizable prefix label** - defaults to "To:" but can be overridden via prefix prop
+- **Conditional QR code scanner** - Appears only when onQrCodeClick handler is provided, using consistent IconButton styling
+- **Smart clear button** - Appears automatically when content is entered
+- **ENS support** - Placeholder indicates support for both address formats
+- **Full validation support** - Complete error handling and state management
```
softeng99, line 23: similarly here

PR #191, opened 2025-09-09
feat: Subheader component

softeng99:
- Implement Subheader component with corresponding documentation


libs/ui-react/src/lib/Components/Subheader/Subheader.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface SubheaderProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> {
+  /**
+   * The title of the subheader.
+   */
+  title: string;
+  /**
+   * The info tooltip of the subheader. When provided, an information icon will be displayed and a tooltip will be shown when hovering over it.
+   */
+  infoTooltip?: React.ReactNode;
```
aammami-ledger, line 14: should this be a string ?

PR #194, opened 2025-09-11
refactor: move composeRefs to ui-core

softeng99:
- moved `composeRefs` utility function to `ui-core`
- refactored build configuration for `ui-core` to be consistent with `utils-shared`


gamegee (approved):
✅ Ok I guess the utils-shared is more recent in term of build configuration, if you make it consistent - good for me

PR #200, opened 2025-09-15
refactor: use slots for the Banner component

softeng99:
- refactored the Banner component to use slots for the actions


libs/ui-react/src/lib/Components/Banner/Banner.tsx
```diff
@@ -6,9 +6,28 @@ import {
...
+
+/**
+ * CloseAction slot component for the Banner. Used to display the close button.
+ */
+const BannerCloseAction = ({ children }: { children: React.ReactNode }) => {
+  return <div className="flex-shrink-0">{children}</div>;
+};
 
 const iconMap = {
   info: <InformationFill className="text-base" />,
```
aammami-ledger, line 33: let's add arial labels here: role="status" (info/success) or role="alert" (warning/error) for screen readers.

libs/ui-react/src/lib/Components/Banner/Banner.tsx
```diff
@@ -6,9 +6,28 @@ import {
...
   DeleteCircleFill,
-  Close,
 } from '../../Symbols';
-import { Button } from '../Button';
+
+/**
+ * PrimaryAction slot component for the Banner. Used to display the primary action button.
+ */
+const BannerPrimaryAction = ({ children }: { children: React.ReactNode }) => {
+  return <div className="flex-shrink-0">{children}</div>;
```
aammami-ledger, line 15: so we need to wrap this a div ? it will add 

maybe we could inject classname to the child ?

libs/ui-react/src/lib/Components/Banner/Banner.tsx
```diff
@@ -65,9 +81,7 @@ export interface BannerProps
  * @param {'info' | 'success' | 'warning' | 'error'} [appearance='info'] - The type of banner which affects color and icon.
  * @param {string} title - The main title of the banner.
  * @param {string} [description] - Optional descriptive text.
- * @param {BannerAction} [primaryAction] - Optional primary action with label and onClick handler.
- * @param {BannerAction} [secondaryAction] - Optional secondary action with label and onClick handler.
- * @param {BannerCloseAction} [closeAction] - Optional close action with onClick handler and optional ariaLabel. Controls the visibility of the close button.
+ * @param {React.ReactNode} [children] - The children of the banner, which can include Banner.PrimaryAction, Banner.SecondaryAction, and Banner.CloseAction slots. Each slot expects a React component (typically a Button) as its child.
```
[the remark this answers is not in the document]
aammami-ledger: Other wise we could add this at the root level: `[&>*]:flex-shrink-0`

libs/ui-react/src/lib/Components/Banner/Banner.tsx
```diff
@@ -85,32 +99,73 @@ export interface BannerProps
...
-    },
+    { appearance = 'info', title, description, className, children, ...props },
     ref,
   ) => {
     const icon = iconMap[appearance];
 
+    const childrenArray = React.Children.toArray(children);
+    const primaryActionSlots = childrenArray.filter(
+      (child) =>
+        React.isValidElement(child) && child.type === BannerPrimaryAction,
```
aammami-ledger, line 140: Currently, slot detection uses strict equality (child.type === BannerPrimaryAction), which may break if components are wrapped in React.memo or HOCs. 
Cursor suggestion: use a displayName convention or Symbol marker to identify slot components more robustly: 
```suggestion
  (child) => React.isValidElement(child) && (child.type as any).displayName === 'BannerPrimaryAction'

```

not a big fan of this as it's not TS safe but I think we should probably take some time to analyse this, for the tooltip it's okay we did not cover this but for components like this one I think there is more possibility that a usecase like this would take place

libs/ui-react/src/lib/Components/Banner/Banner.tsx
aammami-ledger: Other wise we could add this at the root level: [&>*]:flex-shrink-0

libs/ui-react/src/lib/Components/Banner/Banner.tsx
aammami-ledger: Well I ment more like this: 

```
const MemoizedPrimaryAction = React.memo(Banner.PrimaryAction);

<Banner>
  <MemoizedPrimaryAction>
    <Button />
  </MemoizedPrimaryAction>
</Banner>
Here, the slot
``` 

meaning the component itself (Banner.PrimaryAction) has been wrapped.
Now child.type is the `React.memo()` 
So the check fails, even though semantically it’s the same slot.

libs/ui-react/src/lib/Components/Banner/Banner.mdx
```diff
@@ -206,9 +202,18 @@ function ActionableBanner() {
       appearance="warning"
       title="Terms of Service Updated"
       description="Please review and accept the updated terms of service to continue using our platform"
-      primaryAction={{ label: "Accept", onClick: handleAccept }}
-      secondaryAction={{ label: "Review Later", onClick: handleDecline }}
-    />
+    >
```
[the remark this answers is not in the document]
gamegee: Suggestion after checking of [github DS Primer banner](https://primer.style/product/components/banner/)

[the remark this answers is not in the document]
gamegee: Interesting, after investigation on Shadcn [Banner component](https://www.shadcn.io/components/layout/banner) - indeed they use children for composition.

**The shadcn component usage looks like this:**
```tsx
<Banner>
    <BannerIcon icon={CircleAlert} />
    <BannerTitle>Important message</BannerTitle>
    <BannerAction>Learn more</BannerAction>
    <BannerClose />
</Banner>
```

**The shadcn implementation (simplified) looks like this:**
```tsx
const Banner = (...) => {
  // ...
  
  return (<div {...props} className="flex justify-between">
    {children}
  </div>)
}
```

### I observed few interesting points:
1. They are not using compound components (I mean <Banner.MySubComponent />), so it removes the need to create empty components like `PrimaryAction` + the compound typing. 
2. They are not using slots, so the rendered children elements will just be in order. Basically the div containing the sub-elements is just a `flex-row` layout with a `justify-between` class. So no need to go for a `getSlot` and React tree parsing.
3. They also use sub-components for `BannerTitle` and `BannerDescription`, in our current version we are using a mix of properties for `title` and `description` and compound components for actions.

I need to investigate more about **shadcn** and **radix** API design choices to have a clearer global view. At first glance, it seems that if we compare **shadcn** with enterprise design-system like Primer or Polaris they serve different purpose:

### shadcn
  - Flexibility > Constraints
  - Allow engineers to build their own design system
  - Fast iterations with v0, even for "non front-end dev"
### Enterprise DS (polaris, primer)
  - Constraints and Flexibility are important, good tradeoffs is important to find
  - Allow engineers to build end-users product with enterprise guidelines
  - More heavy components, with different interdependencies between them
  
> to conclude, I am in favor of a more consistent pattern and don't have a strong preference between them. I'm curious what you think.

Agreed, we should probably decide to define a guideline to choose between:
- `properties as ReactNode` 
- vs `children compound composition` 
My first thought is we could use both, but I would use `children composition` only when the sub-components are rendered as a simple `children` property, so without using `getSlots`.

Again, just getting onboarded, happy to see your POV 😉 
What do you think @aammami-ledger @softeng99 ?


------
> I like your suggestion about the dismissable prop, although we'll probably need at least an onClose and closeAriaLabel props as well (which starts to look like the existing code that groups these props under a closeAction object)

Yes good point, agreed ✅

PR #202, opened 2025-09-15
feat: Add Figma code connect for new UI components

aammami-ledger:
## code connect 

- subHeader
- sheetBar
- Dialog
- suggestion Keyboard
- AmountInput

## workflow

added code connect workflow that automatically publish and sync updated Figma components to the codebase on pushes to main, failing if any files are unreadable

`paths: src/components/**/*.figma.tsx` → Only triggers if files matching this pattern are changed.


.github/workflows/figma-code-connect.yml
```diff
@@ -0,0 +1,15 @@
+name: Figma Code Connect
+
+on:
+  push:
+    paths:
+      - src/components/**/*.figma.tsx
```
softeng99, line 6: will this path work or should we provide the full path like `libs/**/src/components/**/*.figma.tsx`?

softeng99 (approved):

PR #204, opened 2025-09-17
fix: Dialog animation and accessibility

softeng99:
- add DialogTitle and DialogDescription components for accessibility
- add DialogHeader component as a wrapper for the SheetBar that adds accessibility for the title and description
- remove `iconOnly` prop from the `Button` component

### DIFF
--- a/libs/ui-rnative/src/lib/Components/Button/Button.tsx
+++ b/libs/ui-rnative/src/lib/Components/Button/Button.tsx
@@ -74,7 +74,7 @@ const buttonVariants = cva(
 
 export interface ButtonProps
   extends TouchableOpacityProps,
-    VariantProps<typeof buttonVariants> {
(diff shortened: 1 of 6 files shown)


libs/ui-rnative/src/lib/Components/Button/Button.tsx
```diff
@@ -74,7 +74,7 @@ const buttonVariants = cva(
 
 export interface ButtonProps
   extends TouchableOpacityProps,
-    VariantProps<typeof buttonVariants> {
+    Omit<VariantProps<typeof buttonVariants>, 'iconOnly'> {
```
aammami-ledger, line 77: why omitting `iconOnly` ?

aammami-ledger (approved):
Beautiful PR 🔥

PR #208, opened 2025-09-19
chore/eslint prettier config

gamegee:
## Changes
- **chore: add eslint-plugin-prettier plugin to make it work with eslint**
- **chore: setup eslint-plugin-prettier to extends the prettierrc config and run with eslint**
- **Example of bad commit, formatting checks is catch on the CICD**

## Why ?
This change allow to have only one tool (eslint) to check for quality code checks and stylistics checks.
And prevents the needs for future contributors to install prettier on their machines, because anyway the CI/CD will catch stylistics/formatting issues.

If we agreed on that, you will be able to remove the prettier extension and only rely on eslint.
In your cursor/vscode IDE `User/settings.json`, you will need this config:

```
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": "always"
},
"eslint.validate": [
    "javascriptreact",
    "typescriptreact",
    "typescript",
    "javascript",
    "html"
],
```

Sidenote: I can add this config in the shared `.vscode` of the repository, to ease the future onboarding of engineers. What do you think ?

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,5 +1,6 @@
 import nx from '@nx/eslint-plugin';
 import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';
+import { globalIgnores } from 'eslint/config';
 
 export default [
   ...nx.configs['flat/base'],
@@ -62,4 +63,5 @@ export default [
     rules: {},
   },
   eslintPluginPrettierRecommended,
+  globalIgnores(['**/out-tsc']),
 ];
--- a/nx.json
+++ b/nx.json
@@ -114,6 +114,11 @@
         "^build"
       ]
     },
+    "lint": {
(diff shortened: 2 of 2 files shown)


aammami-ledger (commented):
> Sidenote: I can add this config in the shared .vscode of the repository, to ease the future onboarding of engineers. What do you think ?

yes please do I was gonna suggest that.

libs/ui-react/src/index.ts
```diff
@@ -1,2 +1,8 @@
 export * from './lib/Components';
 export * from './lib/Patterns';
+
+const a = {
+  invalidQuoteFormat: "",
+
+   invalidIndentation: 3,
+};
```
aammami-ledger, line 8: nice, we can revert the commit I guess, point proven.

PR #222, opened 2025-09-26
chore/follow up prettier config singlequote

gamegee:
- **chore(eslint): use singlequote everywhere**
- **chore(eslint): apply autofix**


.prettierrc.mjs
```diff
@@ -1,5 +1,6 @@
 export default {
   singleQuote: true,
+  jsxSingleQuote: true,
```
gamegee, line 3: @aammami-ledger @softeng99 

I would expect to use single-quote everywhere for consistency, this seems reasonable - wdyt ?

softeng99 (approved):
I merged this [PR](https://github.com/LedgerHQ/ldls/pull/219) so please rebase and run lint again before merging

PR #224, opened 2025-09-26
chore(eslint): apply autofix on figma code generation

gamegee:
Signed-off-by: Simon Bruneaud <simon.bruneaud@ledger.fr>

## Desc
- Autofix the formatting using eslint nx task
- As we rely on eslint, we no longer need the dependency `svgr/plugin-prettier`


softeng99 (approved):
awesome 🎉

PR #226, opened 2025-09-29
fix(BaseInput): update styles for active caret and truncate behavior

aammami-ledger:
## Small UI improvements

Truncate both label and value :

image


gamegee (approved):
✅ Nice LGTM

PR #230, opened 2025-09-30
refactor: clean up deprecated class names

aammami-ledger:
## Summary
Replaced Tailwind Typography utilities across the repo with our design-system typography classes and updated internal rules to enforce this.

- Removed deprecated typography class names 
- Updated documentation to reflect changes in class usage for better clarity and consistency.
- Ensured that typography classes are applied correctly to maintain design system standards.
- Updated rules in .cursor/rules/nx-rules.mdc to:
		- Explicitly prohibit Tailwind Typography utilities.
		- Point to libs/design-core/src/utils/create-custom-plugin.ts (createTypographyPlugin) as source of truth.
		- List available classes.


gamegee (approved):
✅

PR #231, opened 2025-09-30
 fix: Rename `iconButton` to `interactiveIcon`

aammami-ledger:
# Rename IconButton → InteractiveIcon

## Why
- Current name `IconButton` suggests a button variant --> causes confusion
- New name `InteractiveIcon` better reflects its purpose as an icon-based interactive element
- Reduces confusion with regular button components

## What Changed
- Renamed all component files from `IconButton.*` to `InteractiveIcon.*`
- Updated all internal references and documentation
- No functional changes, only renaming

## Breaking Change
```diff
- import { IconButton } from '@ledgerhq/ldls-ui-react';
+ import { InteractiveIcon } from '@ledgerhq/ldls-ui-react';

- <IconButton iconType="filled">
+ <InteractiveIcon iconType="filled">
    <DeleteCircleFill />
- </IconButton>
+ </InteractiveIcon>
```

## Migration
Simply rename all instances of `IconButton` to `InteractiveIcon`. Props remain unchanged.

### DIFF
--- a/.nx/version-plans/version-plan-1759240620397.md
+++ b/.nx/version-plans/version-plan-1759240620397.md
@@ -1,0 +1,6 @@
+---
+'@ledgerhq/ldls-ui-rnative': patch
+'@ledgerhq/ldls-ui-react': patch
+---
+
+Breaking change(ui-react): Rename IconButton to InteractiveIcon. Props remain unchanged.


libs/ui-react/.storybook/preview.ts
```diff
@@ -54,7 +54,7 @@ const preview: Preview = {
           'Action',
           [
             'CardButton',
-            'IconButton',
+            'InteractiveIcon',
```
gamegee, line 57: 💯

libs/ui-react/src/lib/Components/InteractiveIcon/InteractiveIcon.mdx
```diff
@@ -1,24 +1,24 @@
 import { Meta, Story, Canvas, Controls } from '@storybook/addon-docs/blocks';
-import * as IconButtonStories from './IconButton.stories';
-import { IconButton } from './IconButton';
+import * as InteractiveIconStories from './InteractiveIcon.stories';
+import { InteractiveIcon } from './InteractiveIcon';
 import { CustomTabs, Tab } from '../../../../.storybook/CustomTabs';
 
-<Meta title="Components/IconButton" of={IconButtonStories} />
+<Meta title='Components/InteractiveIcon' of={InteractiveIconStories} />
```
gamegee, line 6: Later we might decide to choose whether we expose internal components under an `Internal` category

E.g. 
- BaseInput 
- BaseButton
- SheetBar
...

Also a reasonable choice is just to not expose them ?

gamegee (approved):

libs/ui-react/src/lib/Components/InteractiveIcon/InteractiveIcon.mdx
aammami-ledger: I would say not immediately IMHO, but indeed we should at some point decide upon that.

PR #237, opened 2025-10-02
feat/react native checkbox

gamegee:
- **feat(utils-shared): add use-controllable-state from tamagui (RN oriented) + other dependencies of the hook**
- **chore(cursor-rules): add cursor rules for typescript return-type**
- **chore(rn): install @rn-primitives/slot and make the storybook vite build work**
- **feat(rn): add checkbox, types and docs**

### DIFF
--- a/.nx/version-plans/version-plan-1759503867263.md
+++ b/.nx/version-plans/version-plan-1759503867263.md
@@ -1,0 +1,5 @@
+---
+'@ledgerhq/ldls-ui-react': patch
+---
+
+Breaking change(ui-react): Add IconButton and update the Button API
--- a/libs/design-core/src/utils/create-animations-plugin.ts
+++ b/libs/design-core/src/utils/create-animations-plugin.ts
@@ -22,6 +22,38 @@ export function createAnimationsPlugin() {
               },
               to: { opacity: '1', transform: 'translate(-50%, -50%) scale(1)' },
             },
+            slideInFromRight: {
+              from: { transform: 'translateX(10px)', opacity: '0' },
+              to: { transform: 'translateX(0)', opacity: '1' },
+            },
+            slideInFromLeft: {
+              from: { transform: 'translateX(-10px)', opacity: '0' },
+              to: { transform: 'translateX(0)', opacity: '1' },
+            },
+            slideInFromTop: {
+              from: { transform: 'translateY(-10px)', opacity: '0' },
+              to: { transform: 'translateY(0)', opacity: '1' },
+            },
+            slideInFromBottom: {
+              from: { transform: 'translateY(10px)', opacity: '0' },
+              to: { transform: 'translateY(0)', opacity: '1' },
+            },
+            slideOutToRight: {
+              from: { transform: 'translateX(0)', opacity: '1' },
+              to: { transform: 'translateX(10px)', opacity: '0' },
+            },
+            slideOutToLeft: {
+              from: { transform: 'translateX(0)', opacity: '1' },
+              to: { transform: 'translateX(-10px)', opacity: '0' },
+            },
+            slideOutToTop: {
+              from: { transform: 'translateY(0)', opacity: '1' },
+              to: { transform: 'translateY(-10px)', opacity: '0' },
+            },
+            slideOutToBottom: {
+              from: { transform: 'translateY(0)', opacity: '1' },
+              to: { transform: 'translateY(10px)', opacity: '0' },
+            },
             contentHide: {
               from: {
                 opacity: '1',
@@ -38,6 +70,14 @@ export function createAnimationsPlugin() {
             overlayHide: 'overlayHide 250ms cubic-bezier(0.16, 1, 0.3, 1)',
             contentShow: 'contentShow 250ms cubic-bezier(0.16, 1, 0.3, 1)',
             contentHide: 'contentHide 250ms cubic-bezier(0.16, 1, 0.3, 1)',
+            slideInFromRight: 'slideInFromRight 250ms ease-in',
+            slideInFromLeft: 'slideInFromLeft 250ms ease-in',
+            slideInFromTop: 'slideInFromTop 250ms ease-in',
+            slideInFromBottom: 'slideInFromBottom 250ms ease-in',
+            slideOutToRight: 'slideOutToRight 250ms ease-in',
+            slideOutToLeft: 'slideOutToLeft 250ms ease-in',
+            slideOutToTop: 'slideOutToTop 250ms ease-in',
+            slideOutToBottom: 'slideOutToBottom 250ms ease-in',
           },
         },
       },
--- a/libs/ui-react/.storybook/preview.ts
+++ b/libs/ui-react/.storybook/preview.ts
@@ -55,6 +55,7 @@ const preview: Preview = {
           [
             'Button',
             'CardButton',
+            'IconButton',
             'Link',
             'InteractiveIcon',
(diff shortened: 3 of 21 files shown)


libs/utils-shared/src/lib/use-layout-effect/useLayoutEffect.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * On the server, React emits a warning when calling `useLayoutEffect`.
+ * This is because neither `useLayoutEffect` nor `useEffect` run on the server.
+ * We use this safe version which suppresses the warning by replacing it with a noop on the server.
+ *
+ * See: https://reactjs.org/docs/hooks-reference.html#uselayouteffect
+ */
+/* eslint-disable-next-line @typescript-eslint/no-empty-function */
+const useLayoutEffect = globalThis?.document ? React.useLayoutEffect : () => {};
```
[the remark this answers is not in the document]
gamegee: This is to keep the typesafety

libs/ui-rnative/src/lib/Components/Slot/Slot.tsx
```diff
@@ -0,0 +1,205 @@
...
+  type PressableStateCallbackType,
+  type ImageProps as RNImageProps,
+  type ImageStyle as RNImageStyle,
+  type PressableProps as RNPressableProps,
+  type TextProps as RNTextProps,
+  type ViewProps as RNViewProps,
+  type StyleProp,
+} from 'react-native';
+
+const Pressable = React.forwardRef<
```
aammami-ledger, line 17: should we merge this PR given that anything we add in `shared-utils` won't be bundled ? let's maybe temporarily put all this inside `ui-rnative` ? WDYT ? We could add a TODO comment + ticket number to move it once we solve the bundling issue we have.

gamegee: Ok lets move this until we fix the nxjs build process

libs/ui-rnative/src/lib/Components/Checkbox/BaseCheckbox.tsx
```diff
@@ -0,0 +1,167 @@
...
+  SlottablePressableProps
+>(({ asChild, onPress: onPressProp, className, ...props }, ref) => {
+  const { disabled, checked, onCheckedChange, nativeID } = useCheckboxContext();
+
+  function onPress(ev: GestureResponderEvent) {
+    if (disabled) return;
+    const newValue = !checked;
+    onCheckedChange?.(newValue);
+    onPressProp?.(ev);
+  }
```
aammami-ledger, line 113: should we memorise this ?

libs/ui-rnative/src/lib/Components/Checkbox/BaseCheckbox.tsx
```diff
@@ -0,0 +1,167 @@
...
+const baseCheckboxVariants = {
+  trigger: cva(
+    [
+      'size-20 shrink-0 rounded-xs transition-colors',
+      'focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-focus focus-visible:ring-offset-2',
+    ],
+    {
+      variants: {
+        checked: { true: '', false: 'border' },
+        disabled: { true: '', false: '' },
```
aammami-ledger, line 23: do we need this line ?

libs/ui-rnative/src/lib/Components/Checkbox/BaseCheckbox.tsx
gamegee: Yes indeed good catch.

libs/ui-rnative/src/lib/Components/Checkbox/BaseCheckbox.tsx
gamegee: Unfortunately we need it because we are using compoundVariants (line 25).
And CVA can't resolve typesafety properly without this

.cursor/rules/structure-rules.mdc
```diff
@@ -0,0 +1,26 @@
...
+---
+
+# Folder & filenames
+
+When defining a component, the name of the folder and its related files should use PascalCase pattern.
+
+```
+/Button
+  /Button.test.tsx
+  /Button.tsx
```
aammami-ledger, line 14: can you add it should always have a `index.ts` ?

libs/ui-rnative/src/lib/Components/Checkbox/BaseCheckbox.tsx
```diff
@@ -0,0 +1,170 @@
...
+      onPressProp?.(ev);
+    },
+    [disabled, checked, onCheckedChange, onPressProp],
+  );
+
+  const Component = asChild ? Slot.Pressable : Pressable;
+  return (
+    <Component
+      ref={ref}
+      nativeID={nativeID}
```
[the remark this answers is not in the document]
gamegee: In RN you don't really rely on "name" and "value" for choices-input components. Indeed `input` just doesn't exist, from what I've seen from the community, engineers just use `<Pressable>` as an equivalent.

`nativeID` are given as a property, so consumer have to take care of not duplicating it (but its the same in web)

Also for multiple checkbox, we will need a CheckboxGroup component probably, it will make it simpler for devs

libs/ui-rnative/src/lib/Components/Checkbox/BaseCheckbox.tsx
```diff
@@ -0,0 +1,167 @@
...
+  PressableRef,
+  SlottablePressableProps
+>(({ asChild, onPress: onPressProp, className, ...props }, ref) => {
+  const { disabled, checked, onCheckedChange, nativeID } = useCheckboxContext();
+
+  function onPress(ev: GestureResponderEvent) {
+    if (disabled) return;
+    const newValue = !checked;
+    onCheckedChange?.(newValue);
+    onPressProp?.(ev);
```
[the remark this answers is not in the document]
gamegee: Note: The BaseCheckbox is mostly imported from RN-primitive source-code.
I didn't changed it much, expect adding styling and consistent naming

libs/ui-rnative/src/lib/Components/Checkbox/BaseCheckbox.tsx
```diff
@@ -0,0 +1,170 @@
...
+  );
+});
+
+BaseCheckboxTrigger.displayName = 'BaseCheckboxTrigger';
+
+const BaseCheckboxIndicator = React.forwardRef<
+  ViewRef,
+  BaseCheckboxIndicatorProps
+>(({ asChild, forceMount, className, ...props }, ref) => {
+  const { checked, disabled } = useCheckboxContext();
```
aammami-ledger, line 147: [low] we are calling `useCheckboxContext` twice in this file, maybe let's try memorising the values ? or destructure context once in the parent (BaseCheckboxRoot) and pass down only needed props ?

well this is pretty low as it will only has impact if like we have 100 checkboxes in one page  .. 

it's basically a comment to take or leave as you wish

gamegee: Yes its a good point, actually I don't this memoizing the value passed down to the provider will reduce the re-rendering.

But indeed we could use a React.memo() on top of components.

We can take a look at best-in-class DS libs, and see if they use React.memo as a default.

PR #238, opened 2025-10-03
feat/add formcontrol react native

gamegee:
## Description
Add a label and required prop to the checkbox.
Abstract the Label in its own component for now, so it can be re-used for the Switch (and radio?)

- chore(eslint): fix known issue with config of tailwind eslint
- feat(rnative): extend Checkbox API to support label and required props
- feat(rnative): add Label component
- refactor(rnative): remove generated storybook classnames, instead just add text-inherits to components

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,10 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "always"
   },
+  "eslint.workingDirectories": [
+    "./libs/ui-rnative",
+    "./libs/ui-react"
+  ],
   "eslint.validate": [
       "javascriptreact",
       "typescriptreact",
--- a/libs/ui-rnative/src/lib/Components/Button/Button.tsx
+++ b/libs/ui-rnative/src/lib/Components/Button/Button.tsx
@@ -7,22 +7,22 @@ import { Spinner } from '../../Symbols';
 
 const buttonVariants = {
   root: cva(
-    'body-1-semi-bold inline-flex size-fit cursor-pointer flex-row items-center justify-center rounded-full transition-colors',
+    'inline-flex size-fit cursor-pointer flex-row items-center justify-center rounded-full transition-colors body-1-semi-bold',
     {
       variants: {
         appearance: {
           base: 'bg-interactive text-on-interactive active:bg-interactive-pressed',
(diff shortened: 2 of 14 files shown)


libs/ui-rnative/src/lib/Components/Label/Label.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { forwardRef } from 'react';
```
gamegee, line 1: @aammami-ledger I am thinking of doing a `FormControl` folder that will contain few form-related layout components + the label component.

All these would be internal for now, from what I've recently learnt - react-native form components are way less composable and flexible compared to the web due to the lack of "htmlFor" or other utility mecanism.

.vscode/settings.json
```diff
@@ -7,6 +7,10 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "always"
   },
+  "eslint.workingDirectories": [
+    "./libs/ui-rnative",
+    "./libs/ui-react"
+  ],
```
gamegee, line 13: Had issues on vscore/cursor setup with eslint extension that wasn't able to resolve the config properly.
It should be fixed with that :)

libs/ui-rnative/src/lib/Components/Button/Button.tsx
```diff
@@ -5,81 +5,84 @@ import React from 'react';
...
-      {
-        size: 'xs',
-        iconOnly: true,
-        className: 'p-8',
-      },
-      {
-        size: 'sm',
-        iconOnly: true,
-        className: 'p-12',
+const buttonVariants = {
```
aammami-ledger, line 8: seeing this I realised we should also introduce the iconButton in RN and add the baseButton

libs/ui-rnative/src/lib/Components/Label/Label.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { cva } from 'class-variance-authority';
+import { LabelProps } from './types';
+import { Text } from 'react-native';
+import { cn } from '../../utils';
+
+const REQUIRED_SYMBOL = '*';
```
aammami-ledger, line 8: can we validate this on design side ? would like to see this on figma as well and also in the web version

PR #606, opened 2026-04-02
DLS-637/add-dot-symbol-component

gamegee:


aammami-ledger (commented):
Looks goood!

libs/ui-rnative/src/lib/Components/DotSymbol/DotSymbol.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+    rerender(
+      <TestWrapper>
+        <DotSymbol src='https://broken-link.com/404.png' />
+      </TestWrapper>,
+    );
+
+    await waitFor(() => {
+      expect(queryByTestId('dot-symbol-img')).toBeNull();
+    });
+  });
```
aammami-ledger, line 61: this is useless no ? maybe just check it has `bg-muted`

aammami-ledger (approved):

libs/ui-rnative/src/lib/Components/DotSymbol/DotSymbol.test.tsx
gamegee: The `bg-muted` is an implementation detail and could change and therefore break the test for no reason.

Here we are checking that the image is not even displayed if in error, more robust

PR #613, opened 2026-04-07
[M]DLS-629-feature(OptionList): introduce composable OptionList

zel-kass:
## Overview

Display-only selection list designed to live inside a BottomSheet or any container. No trigger, no modal — consumers compose.

- Data-driven grouping via group field on items (auto labels + separators)
- renderItem(item, { selected, disabled }) for full rendering control
- filter / filteredItems props ready for future search implem
- Sub-components: OptionListItem, Leading, Content, ContentRow, Title, Description
- Includes tests, stories, MDX docs, and sandbox examples.

simulator_screenshot_C93FAA37-C360-4DB9-8674-33C157521699


libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,432 @@
...
+      style={StyleSheet.flatten([styles.container, style])}
+      {...props}
+    >
+      <Text style={styles.title}>{title}</Text>
+      {description && <Text style={styles.description}>{description}</Text>}
+    </Box>
+  );
+};
+
+OptionListEmptyState.displayName = 'OptionListEmptyState';
```
aammami-ledger: you need to rebase on `main` we no longer use displayName, see merged [PR](https://github.com/LedgerHQ/lumen/pull/624)

aammami-ledger (commented):
We still need to support the old `SelectTrigger`
so maybe we should expose it as well, maybe call it `SelectInputTrigger`


Image

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,432 @@
...
+
+  if (isGrouped) {
+    return (
+      <Box lx={lx} style={style} ref={ref} {...props}>
+        {groups.map((group, groupIndex) => (
+          <Fragment key={group.label}>
+            {groupIndex > 0 && (
+              <Divider lx={{ marginVertical: 's4', marginHorizontal: 's8' }} />
+            )}
+            <OptionListLabel>{group.label}</OptionListLabel>
```
aammami-ledger:
```suggestion
           group.label && <OptionListLabel>{group.label}</OptionListLabel>
```

to prevent rendering an empty label for groupless items

libs/ui-rnative/src/lib/Components/OptionList/types.ts
```diff
@@ -0,0 +1,99 @@
...
+  children: ReactNode;
+};
+
+export type OptionListContentProps = {
+  /** Render function called for each item. Receives the item data and selection/disabled state. */
+  renderItem: (
+    item: OptionListItemData,
+    state: { selected: boolean; disabled: boolean },
+  ) => ReactNode;
+} & StyledViewProps;
```
aammami-ledger: should omit `children` no ?

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
zel-kass: The grouped branch only runs when isGrouped is true, and label is typed as string hence it is never optional, so the guard wouldn't change anything here. Happy to add it if you prefer though 🙌🏽

libs/ui-rnative/src/lib/Components/OptionList/useOptionList/useOptionListItems.ts
```diff
@@ -0,0 +1,36 @@
+import { useMemo } from 'react';
+import type { OptionListItemData, OptionListItemGroup } from '../types';
+
+const groupByField = (items: OptionListItemData[]): OptionListItemGroup[] =>
+  Array.from(
+    items.reduce<Map<string, OptionListItemData[]>>((acc, item) => {
+      const key = item.group ?? '';
+      return acc.set(key, [...(acc.get(key) ?? []), item]);
```
aammami-ledger: The array spread creates a new array on every iteration, seems unnecessary no ? I would suggest to do this with a `.push(item)` like on the react version



```
export function groupItemsByKey(items: SelectItemData[]): SelectItemGroup[] {
  const order: string[] = [];
  const map: Record<string, SelectItemData[]> = {};
  for (const item of items) {
    const key = item.group ?? '';
    if (!map[key]) {
      order.push(key);
      map[key] = [];
    }
    map[key].push(item);
  }
  return order.map((label) => ({ label, items: map[label] }));
}
```

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
aammami-ledger: yeah but what about items in the group that have no `groupLabel`  ? 

const FOODS: OptionListItemData[] = [
  { value: 'apple', label: 'Apple', group: 'Fruits' },
  { value: 'banana', label: 'Banana', group: 'Fruits' },
  { value: 'orange', label: 'Orange', group: 'Fruits' },
  { value: 'carrot', label: 'Carrot', group: 'Vegetables' },
  { value: 'broccoli', label: 'Broccoli'},
  { value: 'spinach', label: 'Spinach'},
];

zel-kass: Ouh I hadn't considered this case, you're right, adding the suggestion

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+  return (
+    <DisabledProvider value={{ disabled }}>
+      <Pressable
+        ref={ref}
+        lx={lx}
+        style={style}
+        onPress={() => onValueChange(value)}
+        disabled={disabled}
+        accessibilityRole='button'
```
aammami-ledger: should be this "radio" instead ?

libs/ui-rnative/src/lib/Components/OptionList/useOptionList/useOptionListItems.ts
zel-kass: great catch, fixing this right now

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,515 @@
...
+      ]),
+      description: StyleSheet.flatten([
+        t.typographies.body2,
+        { color: t.colors.text.muted },
+      ]),
+    }),
+    [],
+  );
+
+  if (flatItems.length > 0) return null;
```
aammami-ledger, line 417: This condition is always false for grouped data, so the empty state will always render for groups.

Can you add a story to illustrate this empty state as well please

libs/ui-rnative/src/lib/Components/OptionList/types.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** When true, prevents interaction with the entire list. */
+  disabled?: boolean;
+  children: ReactNode;
+};
+
+export type OptionListContentProps = {
+  /** Render function called for each item. Receives the item data and selection/disabled state. */
+  renderItem: (
+    item: OptionListItemData,
+    state: { selected: boolean; disabled: boolean },
```
aammami-ledger: why do we need `disabled` here already got it in `items.disabled`  ? 🤔

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,515 @@
...
+      lx={lx}
+      style={StyleSheet.flatten([styles.leading, style])}
+      {...props}
+    >
+      {children}
+    </Box>
+  );
+};
+
+const OptionListLabel = ({ children }: { children: string }) => (
```
aammami-ledger: can we move the type definition to types ? 
```suggestion
const OptionListLabel = ({ children }: OptionListLabelProps ) => (
```

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,515 @@
...
+            {children}
+            {selected && <Check size={24} color='active' />}
+          </View>
+        )}
+      </Pressable>
+    </DisabledProvider>
+  );
+};
+
+export const OptionListItemTitle = ({
```
aammami-ledger: I have named this `SelectItemText` on the react side, not sure if it's necessary to inline

`SelectItemText` coming from the old radix naming convention so I did not really rename it as part of the mgration

libs/ui-rnative/src/lib/Components/OptionList/types.ts
```diff
@@ -0,0 +1,110 @@
+import { ReactNode } from 'react';
+import {
+  StyledPressableProps,
+  StyledTextProps,
+  StyledViewProps,
+} from '../../../styles';
+
+export type OptionListItemData = {
```
aammami-ledger: I also have `description ` on react side, should align ?

.nx/version-plans/version-plan-1775635930361.md
```diff
@@ -0,0 +1,7 @@
+---
+'@ledgerhq/lumen-ui-rnative': patch
+---
+
+feat(OptionList): add OptionList composable selection list with data-driven grouping, sub-components, tests, stories, and docs.
+
+**Deprecation notice:** OptionList replaces Select. Select will be deprecated in upcoming versions — new selection UIs should use OptionList compositions (e.g. inside BottomSheet).
```
aammami-ledger, line 7: can you add a small migration guide ? it will be useful to give to cursor so hey can do the migration

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,513 @@
...
+        accessibilityState={{ disabled, selected }}
+        {...props}
+      >
+        {({ pressed }) => (
+          <View
+            style={[
+              styles.container,
+              pressed && styles.containerPressed,
+              disabled && styles.containerDisabled,
+            ]}
```
gamegee: [mid]: do these computation inside the stylesheet, so at the end you have:

```tsx
// compute inside
const styles = useItemStyles({ disabled });

<View style={styles.container}>
```

libs/ui-rnative/src/lib/Components/OptionList/OptionList.mdx
```diff
@@ -0,0 +1,302 @@
...
+  const selected = items.find((i) => i.value === value);
+
+  return (
+    <>
+      <OptionListTrigger
+        label='Currency'
+        onPress={() => ref.current?.present()}
+      >
+        {selected && <Text lx={{ color: 'base' }}>{selected.label}</Text>}
+      </OptionListTrigger>
```
gamegee, line 138: [high]: As discussed, now the OptionList is just a list of selectable items.
So you should not even have a trigger, consumers will build their own experience.

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,513 @@
...
+          color: t.colors.text.muted,
+          marginLeft: t.spacings.s8,
+        },
+        disabled && { color: t.colors.text.disabled },
+      ]),
+    }),
+    [disabled, hasValue, hasLabel],
+  );
+
+export const OptionListTrigger = ({
```
gamegee, line 491: [high]: -> Remove it

gamegee (commented):
SonarCloud Quality Gate failed — 65.8% coverage (required >= 80%)

You have untested components like:
- OptionListEmptyState
- OptionListItemLeading
- OptionListItemContentRow

libs/ui-rnative/src/lib/Components/OptionList/types.ts
```diff
@@ -0,0 +1,113 @@
...
```
gamegee: [mid]: suggestion to pass a generic:

```tsx
export type OptionListItemData<TMeta = Record<string, unknown>> = {
...
meta?: TMeta
```

If not already the case in the web, we should do the same

gamegee: This can be very useful for consumers to pass any data:
```tsx
type MyOptionItemData = OptionListItemData<{ additional: string }>
```

aammami-ledger (approved):

gamegee (approved):

libs/ui-rnative/src/lib/Components/OptionList/OptionList.tsx
```diff
@@ -0,0 +1,532 @@
...
+
+  const [selectedValue, setSelectedValue] = useControllableState<string | null>(
+    {
+      prop: value,
+      defaultProp: defaultValue ?? null,
+      onChange: onValueChange,
+    },
+  );
+
+  const { isGrouped, groups, flatItems } = useOptionListItems({ items });
```
gamegee: [mid]: pass tmeta here 

```tsx
useOptionListItems<TMeta>
```

PR #629, opened 2026-04-09
fix(ui-rnative): support "Larger Text" accessibility setting

aquelemiguel:
Closes [DLS-695](https://ledgerhq.atlassian.net/browse/DLS-695).

### Changelog

Implemented in accordance with [ADR 20](https://ledgerhq.atlassian.net/wiki/spaces/DLS/pages/6965526558/ADR+020+Embracing+font+scaling#Current-state).

[DLS-695]: https://ledgerhq.atlassian.net/browse/DLS-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-rnative/src/lib/Components/Banner/Banner.tsx
```diff
@@ -168,7 +172,9 @@ export const Banner = ({
               <Wrap
                 if={isTextChildren(description)}
                 with={(children) => (
-                  <Text style={styles.description}>{children}</Text>
+                  <Text style={styles.description} numberOfLines={5}>
+                    {children}
+                  </Text>
```
aquelemiguel, line 177: Unrelated to font scaling, just took the opportunity to add a limit here.
Laurine and I spotted this yesterday.

zel-kass (approved):
Super clean, gj man 🦾

libs/ui-rnative/src/lib/Components/ListItem/ListItem.tsx
```diff
@@ -29,7 +29,7 @@ const useRootStyles = ({ pressed }: { pressed: boolean }) => {
         {
           flexDirection: 'row',
           alignItems: 'center',
-          height: t.sizes.s64,
+          minHeight: t.sizes.s64,
```
gamegee: [mid]: should we align on web for consistency ? (same for mediabanner, mediacard)

gamegee: Note: After taking a look on storybook, it seems the height of the listItem is now `72px` when it contains a `Spot` - Versus `64px` before

Is that something we expect ?

libs/ui-rnative/src/lib/Components/Tag/Tag.tsx
```diff
@@ -39,15 +39,13 @@ const useStyles = ({
 
       const sizeStyles: Record<
         Size,
-        { height: number; paddingHorizontal: number; paddingVertical: number }
+        { paddingHorizontal: number; paddingVertical: number }
```
gamegee, line 42: [mid]: should we use lineHeight instead of `height` (like other design-system do) - so it scales based on the font-size 👍  and we still have a mapping for `md` and `sm` size.

As the other comment, should we align on react web ?

aquelemiguel: > should we use lineHeight instead of height (like other design-system do) - so it scales based on the font-size 👍 and we still have a mapping for md and sm size.

We're no longer using just height! Vertical size should be calculated by the padding and the respective typography line height, which already changes according to the font size. Did you find any issues?

> As the other comment, should we align on react web ?

I didn't find any fixed height on web!

gamegee: No it seems good ✅ 
Indeed its using the line-height of the inner text element so its fine

gamegee (approved):
✅ LGTM

libs/ui-rnative/src/lib/Components/ListItem/ListItem.tsx
aquelemiguel: Amazing find, potentially solved in https://github.com/LedgerHQ/lumen/pull/629/commits/1cc086c1908cf303ac4f86466dcb9b75f0052094?

gamegee (approved):

PR #634, opened 2026-04-14
feat(DotSymbols): add icon variant for DotSymbol

aammami-ledger:
on Web:

https://github.com/user-attachments/assets/29f2cad7-852f-4bd8-a737-df21554f8bd0

On mobile:
image


libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -1,18 +1,14 @@
...
 import { useEffect, useMemo, useState } from 'react';
-import { MediaImageSize } from '../MediaImage';
-import { SpotSize } from '../Spot';
+import { IconSize } from '../Icon';
 import { DotSymbolPin, DotSymbolProps, DotSymbolSize } from './types';
 
 const dotVariants = cva(
-  'absolute z-10 box-content overflow-hidden border-base-inverted bg-muted',
+  'absolute z-10 box-content flex items-center justify-center overflow-hidden border-base-inverted bg-muted',
   {
```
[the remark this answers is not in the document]
aammami-ledger: cn can resolve overlapping bg-* classes in that single string

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -89,34 +82,65 @@ const getPinOffset = (
...
-  }, [src]);
+  }, [imgSrc]);
+
+  const renderDotContent = () => {
+    if (isIcon) {
+      const { icon: Icon, size: iconSize = 20 } = props;
+      return (
+        <Icon size={dotIconSizeMap[iconSize]} className='text-on-interactive' />
+      );
+    }
```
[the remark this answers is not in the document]
aammami-ledger: TS already handles this

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -89,34 +82,71 @@ const getPinOffset = (
...
```
gamegee: [mid]: can move that into inner file sub-component, instead of a render function
```tsx
const DotContent = () => {

}
```

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -125,16 +155,18 @@ export const DotSymbol = ({
...
+          dotVariants({
+            size,
+            shape,
+            pin,
+            appearance: isIcon ? props.appearance : undefined,
+          }),
         )}
+        style={style}
+      >
+        {renderDotContent()}
```
gamegee:
```tsx
<DotContent size={} ... />
```

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
aammami-ledger: fair! done

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.stories.tsx
aquelemiguel, line 96: unrelated to your PR but i was checking the stories and in this shape showcase, don't we have an extra entry of each here? should be a quick cleanup

Image

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.stories.tsx
```diff
@@ -195,7 +165,107 @@ export const SizeShowcase: Story = {
...
+    type: 'icon',
+    appearance: 'success',
+    icon: ArrowDown,
+    pin: 'bottom-end',
+    size: mediaImageDotSizeMap[48],
+    children: (
+      <MediaImage src={parentSrc} alt='Cardano' size={48} shape='circle' />
+    ),
+  },
+};
```
aquelemiguel: in this story, it's not defaulting to "circle" in the controls

libs/ui-react/src/lib/Components/DotSymbol/types.ts
```diff
@@ -17,6 +25,36 @@ export type DotSymbolProps = {
...
+   * Semantic color of the dot background.
+   */
+  appearance: DotSymbolAppearance;
+  /**
+   * Icon component to render inside the dot.
+   */
+  icon: ComponentType<{ size?: IconSize; className?: string }>;
+  src?: never;
+  alt?: never;
+  imgLoading?: never;
```
aquelemiguel: odd way of doing discriminated unions, is this standard?
usually i default to https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions

libs/ui-rnative/src/lib/Components/DotSymbol/DotSymbol.stories.tsx
aquelemiguel, line 91: some dupes here in the size showcase as well, i think the bottom row should have the square shape (or maybe don't even do two rows, as i think it's not the main purpose of this story)

Image

libs/ui-rnative/src/lib/Components/DotSymbol/types.ts
```diff
@@ -18,6 +27,28 @@ export type DotSymbolProps = {
...
+  /**
+   * Semantic color of the dot background.
+   */
+  appearance: DotSymbolAppearance;
+  /**
+   * Icon component to render inside the dot.
+   */
+  icon: ComponentType<{ size?: IconSize; style?: StyleProp<TextStyle> }>;
+  src?: never;
+  alt?: never;
```
aquelemiguel: i'd argue we could default to `acessibilityLabel` here, as `alt` is more web-related. even if they're alias for eachother, expo-image states:

> The text that's read by the screen reader when the user interacts with the image. Sets the alt tag on web which is used for web crawlers and link traversal. Is an alias for accessibilityLabel.

https://docs.expo.dev/versions/latest/sdk/image/#accessibilitylabel

aquelemiguel (changes requested):

libs/ui-rnative/src/lib/Components/DotSymbol/types.ts
aammami-ledger: no strong opinion here, it was not introduced in this PR though @gamegee wdyt of renaming this ?

libs/ui-react/src/lib/Components/DotSymbol/types.ts
aammami-ledger: The discriminated union already enforces exclusivity on its own; the `?: never` is purely to make variant-specific props reachable on the union so we can strip them from `...rest` in the destructure.

see `DotSymbol.tsx` line `150`

gamegee (commented):
missing version plans

libs/ui-react/src/lib/Components/DotSymbol/types.ts
gamegee: Indeed, I didn't expect to have to force `never` too - the API as well as the implementation start to feel like a code smell.

It makes me think of this conference [from Dan Abramov](https://www.deconstructconf.com/2019/dan-abramov-the-wet-codebase) where he states that sometimes duplication is less harmful than the wrong abstraction.

I am ok to merge it to move forward, however the more I see the component code the more I think that we might split it.

---
**Side-note:** We will need to be careful with breaking changes on media-image and dotSymbol, because both crypto-icons & wallet will share the same peer-deps. So it means Wallet will need to sync multiple libs at the same time.

aammami-ledger:
> missing version plans

added

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -1,18 +1,14 @@
...
 import { cva } from 'class-variance-authority';
 import { useEffect, useMemo, useState } from 'react';
-import { MediaImageSize } from '../MediaImage';
-import { SpotSize } from '../Spot';
+import { IconSize } from '../Icon';
 import { DotSymbolPin, DotSymbolProps, DotSymbolSize } from './types';
 
 const dotVariants = cva(
-  'absolute z-10 box-content overflow-hidden border-base-inverted bg-muted',
+  'absolute z-10 box-content flex items-center justify-center overflow-hidden border-base-inverted bg-muted',
```
gamegee: [low]: `bg-muted` will be overrided if you pass an appearance, maybe it should stand somewhere else in the CVA ?

libs/ui-rnative/src/lib/Components/DotSymbol/types.ts
gamegee: Indeed, both are possible and seems to fullfill the same usage:

I decided to keep alt at first so share the same API - same no strong opinion and anyway the `accessibilityLabel` is extended using `StyledViewProps`

Image

libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
aammami-ledger: yeah and it should! cursor has raised similar point and I decided to keep as is, or I am missing something ?

libs/ui-rnative/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -88,71 +107,133 @@ const useStyles = ({
...
 
   return (
     <Box
       ref={ref}
       lx={lx}
-      style={StyleSheet.flatten([{ position: 'relative' }, style])}
-      accessibilityRole='image'
-      accessibilityLabel={alt}
+      style={StyleSheet.flatten([
+        { position: 'relative', alignSelf: 'flex-start' },
```
gamegee: [mid]: having `alignSelf: flex-start` at the top impact how the component will position on the parent layout. 

This is a common issue on RN because there is no something as `width: fit-content` 😭 (we have RN)

A solution proposed by Claude is to have an extra `Box` to make sure the alignment is only inside the component:
```tsx
<Box
  ref={ref}
  lx={lx}
  style={style}
  accessibilityRole={...}
  accessibilityLabel={...}
  {...rest}
>
  <Box style={{ alignSelf: 'flex-start', position: 'relative' }}>
    {children}
    <Box style={styles.dot}>
      {/* dot content */}
    </Box>
  </Box>
</Box>
```

aammami-ledger:
closing in favour of: https://github.com/LedgerHQ/lumen/pull/647

PR #635, opened 2026-04-14
BREAKING_CHANGE(ui-react/ui-rnative): rename `TriggerButton` to `MediaButton`

aquelemiguel:
Closes [DLS-655](https://ledgerhq.atlassian.net/browse/DLS-655).

### Notable changes

* Renames `TriggerButton` -> `MediaButton`.
* Adds a `hideChevron` prop, self-explanatory.
* Adds a new "block" on the sandbox app to test the component.

### Screenshot

image

[DLS-655]: https://ledgerhq.atlassian.net/browse/DLS-655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gamegee (approved):
I believe you need to update ai-rules too, and maybe the .figma.tsx code-connect ?

Otherwise seems good

aquelemiguel:
> I believe you need to update ai-rules too, and maybe the .figma.tsx code-connect ?

no need on the AI rules but i noticed there's no code-connect for this component
added in the latest commit: https://github.com/LedgerHQ/lumen/pull/635/commits/dd685119fcb8f207da75d8a93bfccd29e5ad4646

libs/ui-react/src/lib/Components/Select/Select.tsx
```diff
@@ -426,15 +426,15 @@ const SelectEmptyState = ({
   </Combobox.Empty>
 );
 
-const SelectTriggerButton = ({
+const SelectMediaButton = ({
```
aammami-ledger, line 429: [high] Not directly related to this PR, but since I was the one who refactored the Select component, I think I made a mistake by introducing the `SelectTriggerButton` subcomponent. We should probably just use `MediaButton` directly instead.

Since this PR already introduces a breaking change in Select, I’m thinking we should simplify things and drop `SelectMediaButton`, and use `MediaButton` directly in the docs and stories.


CC @gamegee FYI

aquelemiguel: while i see the reasoning for doing it all here, i'd prefer this to be a bit more focused
however, i can definitely push this change to another PR today still, got a branch in the works

PR #636, opened 2026-04-14
fix(app-sandbox-rnative): add intl-pluralrules polyfill

aquelemiguel:
Adds the `intl-pluralrules` polyfill to fix that annoying `Intl.PluralRules` compatibility warning we've been having for months now:

> i18next::pluralResolver: Your environment seems not to be Intl API compatible, use an Intl.PluralRules polyfill. Will fallback to the compatibilityJSON v3 format handling

Followed https://stackoverflow.com/a/70865185.
No more issues after testing on my end.

### DIFF
--- a/apps/app-sandbox-rnative/package.json
+++ b/apps/app-sandbox-rnative/package.json
@@ -17,6 +17,7 @@
     "@sbaiahmed1/react-native-blur": "^4.5.5",
     "expo": "~53.0.0",
     "expo-haptics": "~14.1.4",
+    "intl-pluralrules": "^2.0.1",
     "react": "19.0.0",
     "react-native": "0.79.7",
     "react-native-gesture-handler": "^2.28.0",
--- a/apps/app-sandbox-rnative/index.js
+++ b/apps/app-sandbox-rnative/index.js
@@ -1,7 +1,4 @@
-/**
- * @format
- */
-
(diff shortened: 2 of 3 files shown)


apps/app-sandbox-rnative/package.json
```diff
@@ -17,6 +17,7 @@
     "@sbaiahmed1/react-native-blur": "^4.5.5",
     "expo": "~53.0.0",
     "expo-haptics": "~14.1.4",
+    "intl-pluralrules": "^2.0.1",
```
gamegee, line 20: [mid]: can you put that in root dev-deps ?

aquelemiguel: pondered about that as well but ultimately decided against:
1. it's used at runtime, so don't think it fits dev dependencies
2. the polyfill should only really be used in the mobile app cause the issue stems from the RN env, so keen on keeping the dependency in the local package.json

thoughts? good to keep as is?

gamegee: Currently sandbox apps are considered dev only - so it should not make a différence.

Anyway we have a ticket to move all prod deps of app-sanbox to the root.

Its a standard NX convention to setup everything at root in dev deps, so you can reuse it across libs

gamegee (approved):

PR #637, opened 2026-04-14
chart bootstrap libs

gamegee:
[See summary](https://github.com/LedgerHQ/lumen/pull/637#pullrequestreview-4127485173)


libs/ui-rnative-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Rect, Svg, Text as SvgText } from 'react-native-svg';
+import type { Series } from '../internals';
+
+type LineChartProps = {
+  series: Series[];
+  width?: number;
+  height?: number;
+};
+
+export function LineChart({
```
aammami-ledger, line 11: [high] Should be under `src/lib/Components/` to mirror the react one and the other libs

aammami-ledger, line 11: Also folder name should be Uppercased: `Line` maybe `LineChart` even to match the react one ?

libs/ui-react-visualization/src/lib/utils/types.ts
```diff
@@ -0,0 +1,11 @@
...
+  timestamp: number;
+  value: number;
+};
+
+export type Series = {
+  id: string;
+  label: string;
+  data: DataPoint[];
+  color?: string;
+};
```
aammami-ledger, line 11: why not move this tu `utils-shared` as they are shared types ?

libs/ui-react-visualization/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@ledgerhq/lumen-ui-react-visualization",
+  "version": "0.1.0",
```
aammami-ledger, line 3: hmm not `0.0.1` ?

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,31 @@
+import type { Series } from '../../../utils';
+
+type LineChartProps = {
+  series: Series[];
+  width?: number;
+  height?: number;
+};
```
aammami-ledger, line 7: [low] move to `type.ts` ?

libs/ui-rnative-visualization/jest.setup.ts
```diff
@@ -0,0 +1,27 @@
+import { ReactNode } from 'react';
```
aammami-ledger: [low] import type ?

libs/ui-react-visualization/src/lib/utils/types.ts
aammami-ledger: Also, why do we have a `/internals `here? These types are already exported via the barrel, and React doesn’t use an `/internals` structure.

libs/ui-rnative-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,38 @@
...
+  height = 300,
+}: LineChartProps) {
+  return (
+    <View
+      testID='line-chart'
+      accessibilityRole='image'
+      accessibilityLabel='Line chart'
+    >
+      <Svg width={width} height={height}>
+        <Rect width={width} height={height} rx={8} fill='#F3F0FA' />
```
aammami-ledger, line 23: assuming these are just temporary hex values

aammami-ledger (approved):

libs/ui-rnative-visualization/src/lib/Components/LineChart/LineChart.tsx
gamegee: Yes it will be totally overiden

libs/ui-react-visualization/src/lib/utils/types.ts
gamegee: Regarding utils, I will start to keep duplicates as we are currently doing for ui-react vs ui-rnative libs.

I assume that maybe we could group some utils as we implement the charts, but maybe duplicating at first will be harmless.

That being said, this is a real topic we have on API standardization across libs.
Its the kind of healthcheck I would like to run once a month (using a subagent probably)

PR #644, opened 2026-04-20
feat(ui-react): add initialFocus prop to Popover

aammami-ledger:


libs/ui-react/src/lib/Components/Popover/types.ts
```diff
@@ -130,6 +130,24 @@ export type PopoverContentProps = {
...
+   *   Return an element to focus, `true` to use the default behavior, or `false`/`undefined` to do nothing.
+   *
+   * @default false
+   */
+  initialFocus?:
+    | boolean
+    | RefObject<HTMLElement | null>
+    | ((
+        openType: 'mouse' | 'touch' | 'pen' | 'keyboard' | '',
+      ) => boolean | void | HTMLElement | null);
```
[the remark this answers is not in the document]
aammami-ledger: Just mirroring the baseUI [type](https://base-ui.com/react/components/popover) here

gamegee: [low]: Do we need the `function` or could we narrow ? Not even sure of the use-case 🤔 

RefObject can be useful yes 👍 but same we could start by narrowing

gamegee (approved):
✅ LGTM, one concern on narrowing vs fully extending library

libs/ui-react/src/lib/Components/Popover/types.ts
aammami-ledger: Had the same reflection, but then I said it will be odd to take the prop as is same name same everything and narrowing it, people might be confused if they go on baseui docs

aammami-ledger: I will narrow for now

PR #645, opened 2026-04-20
feat(ui-rnative): add DotCount & DotIndicator

aquelemiguel:
Closes the mobile half of [DLS-642](https://ledgerhq.atlassian.net/browse/DLS-642).

[DLS-642]: https://ledgerhq.atlassian.net/browse/DLS-642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-rnative/src/lib/Components/DotCount/DotCount.test.tsx
```diff
@@ -0,0 +1,67 @@
+import { describe, it, expect } from '@jest/globals';
+import { ledgerLiveThemes } from '@ledgerhq/lumen-design-core';
```
gamegee, line 2: [mid]: might add tests case to reach 80% sonarqube test coverage

libs/ui-rnative/src/lib/Components/DotCount/DotCount.tsx
```diff
@@ -0,0 +1,112 @@
...
+export function DotCount({
+  value,
+  size,
+  max = 99,
+  appearance = 'base',
+  disabled: disabledProp = false,
+  lx,
+  style,
+  children,
+}: DotCountProps) {
```
gamegee, line 19: [high]: should props drill other props (including ref, ...)

libs/ui-rnative/src/lib/Components/DotIndicator/DotIndicator.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { DotIndicatorProps } from './types';
+
+export function DotIndicator({
+  appearance = 'base',
+  disabled: disabledProp = false,
+  lx,
+  style,
+  children,
+  accessibilityLabel,
+}: DotIndicatorProps) {
```
gamegee, line 17: [high]: should props drill other props (including ref, ...)

libs/ui-rnative/src/lib/Components/DotIndicator/DotIndicator.tsx
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  const styles = useStyles({
+    appearance,
+    disabled,
+    pinned: !!children,
+  });
+
+  return (
+    <Box>
```
gamegee: [mid]: Wrapper architecture is inverted,  Consumer props (lx, style) are applied to the inner dot element instead of the outer wrapper. Compare with DotSymbol

libs/ui-rnative/src/lib/Components/DotCount/DotCount.tsx
```diff
@@ -0,0 +1,112 @@
...
+    size,
+    appearance,
+    disabled,
+    pinned: !!children,
+  });
+
+  const cappedMax = Math.min(max > 0 ? max : 99, 99);
+
+  return (
+    <Box>
```
gamegee: [mid]: Wrapper architecture is inverted,  Consumer props (lx, style) are applied to the inner dot element instead of the outer wrapper. Compare with DotSymbol

libs/ui-rnative/src/lib/Components/DotCount/DotCount.tsx
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  const styles = useStyles({
+    size,
+    appearance,
+    disabled,
+    pinned: !!children,
+  });
+
+  const cappedMax = Math.min(max > 0 ? max : 99, 99);
```
gamegee: [low]: `max <= 0` silently replaced with 99

libs/ui-rnative/src/lib/Components/DotIndicator/types.ts
```diff
@@ -0,0 +1,20 @@
+import type { ReactNode } from 'react';
+import type { StyledViewProps } from '../../../styles';
+
+export type DotIndicatorProps = {
+  /**
+   * The appearance of the dot indicator.
+   * @default base
+   */
+  appearance?: 'base' | 'red';
```
aammami-ledger, line 14: [high] would go with `error` not `red` honestly 
But at the same time this is not really an error state :/ 

We have never used the primitive colors as prop names, and I do think we should NOT as this is not scalable, it could one day become any other color.. 

Maybe we can find a better naming ? maybe 'negative' like Coinbase ? 

Image

libs/ui-rnative/src/lib/Components/DotCount/DotCount.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const Base: Story = {
+  args: {
+    value: 3,
+    size: 'md',
+    appearance: 'base',
+  },
+};
+
+export const SizeShowcase: Story = {
```
aammami-ledger, line 53: [mid] missing an exemple of when it takes children.

libs/ui-rnative/src/lib/Components/DotIndicator/DotIndicator.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+      control: 'radio',
+      options: ['base', 'red'],
+    },
+  },
+} satisfies Meta<typeof DotIndicator>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Base: Story = {
```
aammami-ledger, line 33: [mid] same comment on the DotCount, pretty sure this one will probably never be used stand alone 🤔 
We should do as for [Dotsymbol](https://lumen-ldls.vercel.app/?path=/docs/communication-dotsymbol--docs)

aammami-ledger (commented):
What did we decide about reusing the DotIndicator on [Avatar](https://lumen-ldls.vercel.app/?path=/docs/communication-avatar--docs#notification-indicator), I fail de remember

libs/ui-rnative/src/lib/Components/DotCount/DotCount.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { Box, Text } from '../Utility';
+import type { DotCountProps } from './types';
+
+export function DotCount({
+  value,
+  size = 'md',
+  max = 99,
+  appearance = 'base',
+  disabled: disabledProp = false,
+  lx,
```
aammami-ledger: [low] I know we default this to `{}` in all other components maybe we should keep this pattern. Same for DotIndicator
```suggestion
 lx = {}
```

libs/ui-rnative/src/lib/Components/DotIndicator/DotIndicator.tsx
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  const styles = useStyles({
+    appearance,
+    disabled,
+    pinned: !!children,
+  });
+
+  return (
+    <Box ref={ref} lx={lx} style={style} {...props}>
```
aammami-ledger: [mid] probably missing the `position: 'relative'` in order for this to work when it has children (see DotSymbol)
```suggestion
<Box ref={ref} lx={lx} style={StyleSheet.flatten([children && { position: 'relative' }, style])} {...props}>
```

libs/ui-rnative/src/lib/Components/DotCount/DotCount.tsx
```diff
@@ -0,0 +1,114 @@
...
+    appearance,
+    disabled,
+    pinned: !!children,
+  });
+
+  const cappedMax = Math.max(1, Math.min(max, 99));
+
+  return (
+    <Box ref={ref} lx={lx} style={style} {...props}>
+      <Box style={styles.container} pointerEvents='none'>
```
aammami-ledger: [low] missing `accessibilityRole`

libs/ui-rnative/src/lib/Components/DotIndicator/types.ts
aquelemiguel: i completely agree with your points, it itched me as well
found it justified because button uses the `red` appearance but i think it's wrong as well
went with `negative`, makes most sense and should be our new pattern, wdyt?

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -112,9 +106,6 @@ export const Avatar = ({
       accessibilityLabel={accessibilityLabel}
       {...props}
     >
-      {showNotification && (
-        <View style={styles.notification} accessible={false} />
```
aammami-ledger, line 116: we are loosing the 	`accessible={false}` no ? is there a way to put it back on the Dot Indicator 🤔

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -132,4 +123,14 @@ export const Avatar = ({
       )}
     </Box>
   );
+
+  if (showNotification) {
+    return (
+      <DotIndicator size={dotSizeMap[size]} appearance='red'>
+        {avatarContent}
+      </DotIndicator>
```
aammami-ledger: i am a bit concern this will cause some issues to consumers  as Avatar tree structure changes here 🤔 as now it's wrapped in `DotIndicator`
for the `lx`,`ref` and `styles` makes sense that they are passed to  avatarContent, but maybe `test ID` should be at the `DotIndicator` level ? 

not sure honestly maybe it's okay

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: good shout, i can make it so the DotIndicator's `accessible` prop is only true if we pass it a label
in the avatar's case we don't pass anything to it so all good

aquelemiguel: so in DotIndicator.tsx...

```tsx
<Box
    style={styles.container}
    accessibilityRole='image'
    accessibilityLabel={accessibilityLabel}
    accessible={!!accessibilityLabel}   // <--
    pointerEvents='none'
/>
```

libs/ui-rnative/src/lib/Components/DotCount/DotCount.tsx
```diff
@@ -0,0 +1,127 @@
...
+      lx={lx}
+      style={StyleSheet.flatten([
+        children ? { position: 'relative' } : undefined,
+        style,
+      ])}
+      {...props}
+    >
+      <Box
+        style={styles.container}
+        accessibilityRole='image'
```
aammami-ledger, line 46: should do as in `DotIndicator` and forward `accessibilityLabel` here instead of outer box ?

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: our DotIndicator wrapper doesn't do much, just `position: relative`, so layout impact is minimal
however the i agree with your concern about `testId`, can adjust just to be safe

aammami-ledger (approved):

libs/ui-rnative/src/lib/Components/DotCount/DotCount.tsx
aquelemiguel: yup, same pattern

PR #648, opened 2026-04-20
fix(Select): normalize keyboard selection values & add generic TMeta support

zel-kass:
## Overview

- Fix onValueChange returning full objects on keyboard selection: base-ui's Combobox resolves keyboard-selected items from its internal items array, yielding SelectItemData objects instead of plain strings. A resolveValue normalizer + isItemEqualToValue override ensure onValueChange always emits a string | null.
- Fix arrow-key navigation when SelectSearch is absent: Combobox requires an Input element to handle arrow keys. A hidden SelectFallbackInput is now rendered inside SelectContent when no SelectSearch is mounted.
- Align SelectItemData generics with OptionListItemData: Introduced MetaShape and replaced the unconstrained Meta default with TMeta extends MetaShape = MetaShape, threaded through SelectItemGroup, SelectProps, SelectListProps, useSelectItems, and groupItemsByKey — matching the React Native OptionList pattern.

### DIFF
--- a/libs/ui-react/src/lib/Components/Select/Select.test.tsx
+++ b/libs/ui-react/src/lib/Components/Select/Select.test.tsx
@@ -121,6 +121,53 @@ describe('Select', () => {
     expect(handleChange).toHaveBeenCalledWith('opt1');
   });
 
+  it('normalizes object values to strings in onValueChange', () => {
+    const handleChange = vi.fn();
+    const { rerender } = render(
+      <Select items={options} onValueChange={handleChange}>
+        <SelectTrigger label='Label' />
+        <SelectContent>
+          <SelectList
+            renderItem={(item) => (
+              <SelectItem key={item.value} value={item.value}>
+                <SelectItemText>{item.label}</SelectItemText>
+              </SelectItem>
+            )}
+          />
+        </SelectContent>
+      </Select>,
+    );
+
+    fireEvent.click(screen.getByRole('combobox'));
+    fireEvent.click(screen.getByText('Option 1'));
+
+    expect(handleChange).toHaveBeenCalledWith('opt1');
+    expect(typeof handleChange.mock.calls[0][0]).toBe('string');
+
+    handleChange.mockClear();
+
+    rerender(
+      <Select items={options} onValueChange={handleChange} value={null}>
+        <SelectTrigger label='Label' />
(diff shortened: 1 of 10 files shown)


libs/ui-react/src/lib/Components/Select/Select.test.tsx
```diff
@@ -121,6 +121,53 @@ describe('Select', () => {
...
+                <SelectItemText>{item.label}</SelectItemText>
+              </SelectItem>
+            )}
+          />
+        </SelectContent>
+      </Select>,
+    );
+
+    fireEvent.click(screen.getByRole('combobox'));
+    fireEvent.click(screen.getByText('Option 1'));
```
aammami-ledger: maybe add a test for the keyboard selection ?

aammami-ledger (approved):

libs/ui-react/src/lib/Components/Select/utils/groupItems.ts
```diff
@@ -1,13 +1,33 @@
-import type { SelectItemData, SelectItemGroup } from '../types';
+import type { MetaShape, SelectItemData, SelectItemGroup } from '../types';
+
+/**
+ * base-ui's Combobox calls onValueChange with either a string (click on
+ * Combobox.Item) or a full SelectItemData object (keyboard selection, which
+ * resolves from the `items` array). This normalizes both to a plain string.
+ */
+export const resolveValue = (v: unknown): string | null => {
```
gamegee: [mid]: rename `value`

libs/ui-react/src/lib/Components/Select/Select.tsx
```diff
@@ -41,7 +43,7 @@ function Select({
   name,
   required,
   children,
-}: SelectProps) {
+}: Readonly<SelectProps<TMeta>>) {
```
gamegee: [low]: we should not need `Readonly` here, react is already a framework that prevent props mutation at its core principles.

aammami-ledger: it's SonarQube that was failing due to this, so I said no harm in adding it.

zel-kass: just removed it 👀

PR #661, opened 2026-04-28
Support cssVar in addition of tailwind tokens

gamegee:


libs/design-core/src/lib/themes/css/primitives-css.ts
```diff
@@ -684,7 +684,7 @@ export const tokens: Record<string, Record<string, string | number>> = {
     '--stroke-1': '1px',
     '--stroke-2': '2px',
     '--stroke-4': '4px',
-    '--stroke-05': '0.5px',
+    '--stroke-0,5': '0.5px',
```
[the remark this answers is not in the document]
aquelemiguel: ^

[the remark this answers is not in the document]
gamegee: Indeed, but this is autogenerated from Figma API - I will check with Laurine

zel-kass (approved):
lvfgtm

libs/design-core/src/lib/cssVar/cssVar.ts
```diff
@@ -0,0 +1,17 @@
+import type { allBrandsCSSTheme, primitivesTheme } from '../themes/css/index';
+
+type PrimitiveKeys = keyof (typeof primitivesTheme)[':root'];
+type SemanticKeys = keyof (typeof allBrandsCSSTheme)['.ledger-live'];
```
aammami-ledger, line 4: [high] so we only get the ones of the `ledger-live` preset ? 🤔

gamegee: yes, for all theme we have the same keys, no differences - only the values differs

gamegee: This is also what we do for react-native lib types. Didn't find a better way for now.
We could probably factorize the theme object at some point to only have the accent color that is customized, that's the only one ?

aammami-ledger (approved):
LGTM! added the `chromatic` tag so we can check no regression after this small refacto.

PR #662, opened 2026-04-28
feat(ui-react): add NavBarDescription and NavBarLeading

aammami-ledger:
image

### DIFF
--- a/libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
+++ b/libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
@@ -4,6 +4,8 @@ import {
   NavBar,
   NavBarBackButton,
   NavBarCoinCapsule,
+  NavBarDescription,
+  NavBarLeading,
   NavBarTitle,
   NavBarTrailing,
 } from './NavBar';
@@ -37,7 +39,9 @@ NavBar is a navigation component that provides a consistent header for pages wit
 
 - **NavBar**: The root container that provides horizontal layout with proper spacing
 - **NavBarBackButton**: An icon button with arrow-left icon for back navigation
+- **NavBarLeading**: Container for the main content area (title + optional description side-by-side). Mirrors `NavBarTrailing` on the left side
 - **NavBarTitle**: The page title text, automatically truncates when content overflows
+- **NavBarDescription**: Secondary descriptive text displayed next to the title inside `NavBarLeading`
 - **NavBarCoinCapsule**: A pill-shaped container for displaying cryptocurrency icon and ticker (e.g., for asset detail pages)
 - **NavBarTrailing**: Container for right-aligned action elements like IconButtons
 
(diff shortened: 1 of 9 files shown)


libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
```diff
@@ -61,6 +65,13 @@ Multiple action buttons can be placed in the trailing area:
 
 <Canvas of={NavBarStories.WithMultipleTrailingActions} />
 
+### With Description
+
+Use `NavBarLeading` to group a title and description side-by-side with a 16px gap.
+Pass `className="flex-none"` to `NavBarTitle` so it does not expand and push the description away.
```
gamegee: Should it be the default ? 🤔 
Are there other way to handle it ? Because it feels like the component will be broken without it

libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
```diff
@@ -121,6 +132,43 @@ function MyPage() {
...
+      </NavBarLeading>
+      <NavBarTrailing>
+        <IconButton
+          appearance='gray'
+          size='sm'
+          icon={Settings}
+          aria-label='Settings'
+        />
+      </NavBarTrailing>
+    </NavBar>
```
gamegee, line 164: [mid] on the same comment, if we would have the BackButton inside the Leading, you could just have a Navbar with `spaceBetween` that would handle the spacing correctly ?

2nd solution - Or maybe other solution have the flex-1 on the `NavbarLeading` and not the `NavbarTitle` ?

libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
aammami-ledger: yeah sorry I have not comited my latest changes, this was fully AI I updated this part

libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
aammami-ledger: yes indeed same as comment above

aammami-ledger: The thing is there only a `4px` gap between arrow and title :/  and we have one of `16px` between title + description

aammami-ledger: What I can do is handle this with some margins not gap, but not optimal indeed

libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
aammami-ledger: image

should be good now, this what Laurine wants.

libs/ui-react/src/lib/Components/NavBar/NavBar.mdx
gamegee: Ah yes indeed, lets keep it like this then

gamegee (approved):

PR #665, opened 2026-04-28
[D]DLS-706-feat(visualisation): Add Point subcomponent

zel-kass:
## Overview

Introduce Point subcomponent, that allows to project a point onto a graph, supporting a label as a string or custom component.

---
Screenshot 2026-04-28 at 16 54 19

### DIFF
--- a/libs/ui-react-visualization/src/lib/Components/CartesianChart/CartesianChart.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/CartesianChart/CartesianChart.tsx
@@ -1,64 +1,13 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import type { ChartInset } from '../../utils/types';
 
 import { CartesianChartProvider, useBuildChartContext } from './context';
 import type { CartesianChartProps } from './types';
-
-const DEFAULT_HEIGHT = 160;
-
-/**
- * Internal buffer that prevents SVG content (labels, points, ticks) from being
- * clipped at the edge of the container. Compensated by negative margins on the
- * wrapper so the chart's visual footprint stays unchanged.
- */
-const OVERFLOW_BUFFER: ChartInset = {
-  top: 30,
-  right: 20,
-  bottom: 30,
-  left: 20,
-};
-const OVERFLOW_NEGATIVE_MARGIN = {
-  marginTop: -OVERFLOW_BUFFER.top,
-  marginRight: -OVERFLOW_BUFFER.right,
-  marginBottom: -OVERFLOW_BUFFER.bottom,
-  marginLeft: -OVERFLOW_BUFFER.left,
-};
-const ZERO_PADDING: ChartInset = { top: 0, right: 0, bottom: 0, left: 0 };
-
-const resolveInset = (inset: CartesianChartProps['inset']): ChartInset => {
-  let consumer: ChartInset;
-  if (inset === undefined) {
-    consumer = ZERO_PADDING;
-  } else if (typeof inset === 'number') {
-    consumer = { top: inset, right: inset, bottom: inset, left: inset };
-  } else {
-    consumer = {
-      top: inset.top ?? 0,
-      right: inset.right ?? 0,
-      bottom: inset.bottom ?? 0,
-      left: inset.left ?? 0,
-    };
-  }
-
-  return {
-    top: consumer.top + OVERFLOW_BUFFER.top,
-    right: consumer.right + OVERFLOW_BUFFER.right,
-    bottom: consumer.bottom + OVERFLOW_BUFFER.bottom,
-    left: consumer.left + OVERFLOW_BUFFER.left,
-  };
-};
-
-const resolveAxisPadding = (
-  padding: CartesianChartProps['axisPadding'],
-): ChartInset => {
-  if (padding === undefined) return ZERO_PADDING;
-  return {
-    top: padding.top ?? 0,
-    right: padding.right ?? 0,
-    bottom: padding.bottom ?? 0,
-    left: padding.left ?? 0,
-  };
-};
+import {
+  DEFAULT_HEIGHT,
+  OVERFLOW_NEGATIVE_MARGIN,
+  resolveAxisPadding,
+  resolveInset,
+} from './utils';
 
 export function CartesianChart({
   series,
--- a/libs/ui-react-visualization/src/lib/Components/CartesianChart/CartesianChart.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/CartesianChart/CartesianChart.tsx
@@ -6,23 +6,28 @@ import type { CartesianChartProps } from './types';
 
 const DEFAULT_HEIGHT = 160;
 const DEFAULT_INSET: ChartInset = {
-  top: 8,
-  right: 0,
-  bottom: 0,
-  left: 0,
+  top: 30,
+  right: 20,
+  bottom: 30,
+  left: 20,
 };
 const ZERO_PADDING: ChartInset = { top: 0, right: 0, bottom: 0, left: 0 };
 
 const resolveInset = (inset: CartesianChartProps['inset']): ChartInset => {
   if (inset === undefined) return DEFAULT_INSET;
   if (typeof inset === 'number') {
-    return { top: inset, right: inset, bottom: inset, left: inset };
+    return {
+      top: inset + DEFAULT_INSET.top,
+      right: inset + DEFAULT_INSET.right,
+      bottom: inset + DEFAULT_INSET.bottom,
+      left: inset + DEFAULT_INSET.left,
+    };
   }
   return {
-    top: inset.top ?? DEFAULT_INSET.top,
-    right: inset.right ?? DEFAULT_INSET.right,
-    bottom: inset.bottom ?? DEFAULT_INSET.bottom,
-    left: inset.left ?? DEFAULT_INSET.left,
+    top: (inset.top ?? 0) + DEFAULT_INSET.top,
+    right: (inset.right ?? 0) + DEFAULT_INSET.right,
+    bottom: (inset.bottom ?? 0) + DEFAULT_INSET.bottom,
+    left: (inset.left ?? 0) + DEFAULT_INSET.left,
   };
 };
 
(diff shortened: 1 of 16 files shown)


libs/ui-react-visualization/src/lib/Components/CartesianChart/CartesianChart.tsx
```diff
@@ -5,24 +5,46 @@ import { CartesianChartProvider, useBuildChartContext } from './context';
...
+};
+const OVERFLOW_NEGATIVE_MARGIN = {
+  marginTop: -OVERFLOW_BUFFER.top,
+  marginRight: -OVERFLOW_BUFFER.right,
+  marginBottom: -OVERFLOW_BUFFER.bottom,
+  marginLeft: -OVERFLOW_BUFFER.left,
 };
 const ZERO_PADDING: ChartInset = { top: 0, right: 0, bottom: 0, left: 0 };
 
 const resolveInset = (inset: CartesianChartProps['inset']): ChartInset => {
```
gamegee: [low]: possibly at some point, move that into utils + the constants

libs/ui-react-visualization/src/lib/Components/Point/Point.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const fill = color ?? cssVar('var(--background-muted-strong)');
+
+  const pixel = useMemo(() => {
+    if (!xScale || !yScale) return undefined;
+    return projectPoint(dataX, dataY, xScale, yScale);
+  }, [dataX, dataY, xScale, yScale]);
+
+  if (!pixel || !isWithinBounds(pixel.x, pixel.y, drawingArea)) {
+    return null;
+  }
```
gamegee, line 66: [low]: this part fail silently, maybe at somepoint we could have logger for end consumers:

```tsx
lumenLogger.warn({ 
  component: 'Point'
  message: 'Point x:${dataX} y:${dataY} is outside of bounds' 
})
```

And logger would only log if debug=true:
```tsx
<ThemeProvider debug={true} />
```

---
Just long term idea, could implement and standardize it later

libs/ui-react-visualization/src/lib/Components/Point/Point.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+    expect(label.textContent).toBe('#2');
+  });
+
+  it('renders labelComponent when provided', () => {
+    const { getByTestId, queryByTestId } = renderInChart(
+      <Point
+        dataX={2}
+        dataY={30}
+        label='ignored'
+        labelComponent={<text data-testid='custom-label'>Custom</text>}
```
gamegee: [mid]: Have you tested with non-svg elements, like div, span ?

libs/ui-react-visualization/src/lib/Components/Point/types.ts
```diff
@@ -0,0 +1,53 @@
...
+   * Diameter of the point circle in pixels.
+   * Internally converted to an SVG radius (`size / 2`).
+   * @default 10
+   */
+  size?: number;
+  /**
+   * When `true`, an arrow is displayed pointing from the point to the label.
+   * @default true
+   */
+  showArrow?: boolean;
```
gamegee: [low]: more accurate `showLabelArrow` ? 
Because this will not be displayed if you don't have label

libs/ui-react-visualization/src/lib/Components/Point/types.ts
```diff
@@ -0,0 +1,53 @@
...
+   * A string is rendered directly; a function receives `dataX` and returns
+   * the label string, enabling index-based formatting.
+   */
+  label?: string | ((dataIndex: number) => string);
+  /**
+   * Custom SVG element rendered instead of the default text label.
+   * Automatically translated so that `(0, 0)` is the label anchor point
+   * (horizontally centered on the point, vertically offset like the built-in label).
+   */
+  labelComponent?: ReactNode;
```
gamegee: [mid]: For the labelComponent, we should inject the props like this:

```tsx
<Point 
  label="coucou" 
  labelComponent={({ children, ...props }) => 
    <MyLabelComponent {...props}>
      {children}
    </MyLabelComponent>
 }
```

The children should be the resolved label passed from the parent, meaning in your JSX part of the point you should have:

```tsx
const LabelComponent = LabelComponent
const hasLabelComponent = Boolean(LabelComponent)

...
{hasLabelComponent && <LabelComponent x={x} y={y}>{resolvedLabel}</LabelComponent>}
```

That how Coinbase is doing, and they provide a `<ChartText/> sub-component to do so.
See example https://cds.coinbase.com/components/charts/Point/?tab=examples#custom-label-position
See code https://github.com/coinbase/cds/blob/f804872ceb5ecd483aeb2086a35cf7fd86e1bb64/packages/web-visualization/src/chart/point/Point.tsx#L147


---
That being said, I believe we can't really render `span` or `div` inside the svg as custom element.
So either we 
- (1) expose sub-component like ChartText
- (2) already position the custom label wrapper and pass resolved label + provide clear example on how to use it.
- mix both ?

libs/ui-react-visualization/src/lib/Components/Point/types.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Fill color of the point circle.
+   */
+  color?: string;
+  /**
+   * Text label displayed near the point.
+   * A string is rendered directly; a function receives `dataX` and returns
+   * the label string, enabling index-based formatting.
+   */
+  label?: string | ((dataIndex: number) => string);
```
gamegee, line 40: 👍

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.stories.tsx
```diff
@@ -237,3 +238,24 @@ export const WithAreaMultipleSeries: Story = {
     },
   },
 };
+
+export const WithPoints: Story = {
```
gamegee: Add one with `LabelComponent`

libs/ui-react-visualization/src/lib/Components/CartesianChart/CartesianChart.tsx
zel-kass: love that, did it for the point, doing it right now for CartesianChart 🙌🏽

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.stories.tsx
gamegee: Maybe start to split stories between LineChart, Point ? 
And start to match Figma specs as stories (gray, green & red, ...)

libs/ui-react-visualization/src/lib/Components/Point/Point.tsx
zel-kass: Indeed this would be a great idea, let's keep it in mind in the charts roadmap

libs/ui-react-visualization/src/lib/Components/Point/Point.test.tsx
zel-kass: Since HTML elements can't be placed inside svg, I didn't test it cause this would just validate a misuse scenario. Do you think we should still test ?

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.stories.tsx
```diff
@@ -259,3 +259,78 @@ export const WithPoints: Story = {
...
+            />
+            <PointLabel
+              x={x}
+              y={y}
+              style={{ fill: '#fff', fontSize: 11, fontWeight: 500 }}
+            >
+              {children}
+            </PointLabel>
+          </g>
+        )}
```
gamegee: [mid]: Do we want to allow that kind of example with low level-primitives ? 
I would stick only with PointLabel one.

Maybe a useful example would be with 2 points label - one below the other to display multiple information ? wdyt ?

```tsx
<PointLabel x={x} y={y}>
  <PointLabel>First text</PointLabel>
  <PointLabel>2nd text</PointLabel>
</PointLabel>
```

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.stories.tsx
```diff
@@ -259,3 +259,78 @@ export const WithPoints: Story = {
...
+              fill='#C24244'
+            />
+            <polygon
+              points={`${x - 4},${y} ${x},${y - 5} ${x + 4},${y}`}
+              fill='#C24244'
+            />
+            <PointLabel
+              x={x}
+              y={y + 14}
+              style={{ fill: '#fff', fontSize: 11, fontWeight: 600 }}
```
gamegee: [mid]: don't use pure hexa but `cssVar` instead -> The AI will source the examples, so its important to enforce consistency and constraints

gamegee (approved):

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.stories.tsx
```diff
@@ -1,9 +1,11 @@
+import { cssVar } from '@ledgerhq/lumen-design-core';
 import type { Meta, StoryObj } from '@storybook/react-vite';
 
 import { StoryDecorator } from '../../../../.storybook/StoryDecorator';
+import { Point, PointLabel } from '../Point/Point';
```
aammami-ledger:
```suggestion
import { Point, PointLabel } from '../Point';
```

libs/ui-react-visualization/src/lib/Components/Point/types.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * X coordinate in data space (index or explicit value).
+   */
+  dataX: number;
+  /**
+   * Y coordinate in data space.
+   */
+  dataY: number;
+  /**
+   * Fill color of the point circle.
```
aammami-ledger, line 21: can you add some guidance about using the `cssVar` utils to set the color, maybe add an exemple

libs/ui-react-visualization/src/lib/Components/Point/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+export const STROKE_WIDTH = 2;
+export const ARROW_WIDTH = 6;
+export const ARROW_HEIGHT = 4;
+export const GAP = 4;
+export const LABEL_FONT_SIZE = 10;
+
+export const isWithinBounds = (
+  px: number,
+  py: number,
+  area: { x: number; y: number; width: number; height: number },
```
aammami-ledger: maybe use `DrawingArea` ?
```suggestion
  area: DrawingArea,
```

aammami-ledger (approved):
Nice one 👏

PR #673, opened 2026-04-29
fix/table-cell-overflow-issue

gamegee:


zel-kass (approved):

libs/ui-react/src/lib/Components/Table/Table.tsx
```diff
@@ -277,7 +277,7 @@ export const TableCell = ({
   );
 };
 
-const cellContentVariants = cva('flex items-center gap-12 truncate', {
+const cellContentVariants = cva('flex min-w-0 items-center gap-12', {
```
aammami-ledger, line 280: not sure i understand how this fixes the overflow 🤔

gamegee: Its a common workaround with flex layouts -->

In a flex layout, flex children have an implicit `min-width: auto` by default. This means a flex child will never shrink below its content size, it pushes its parent wider instead of allowing text to overflow and truncate.

`min-w-0` overrides that default to `min-width: 0`, which tells the flex child: "you're allowed to be smaller than your content." This is what actually enables the inner truncate on the text divs to work, without it, the text container would just grow and push the layout out instead of truncating.

PR #675, opened 2026-04-30
feat(MediaImage): enhance fallback and loading states

aammami-ledger:
image


libs/ui-react/src/lib/Components/CryptoIcon/CryptoIcon.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default meta;
+
+export const CryptoIcons: StoryObj = {
+  render: () => (
+    <iframe
+      src={CRYPTO_ICONS_URL}
+      title='Crypto Icons'
+      style={{ width: '100%', height: '100vh', border: 'none' }}
+    />
```
aammami-ledger: Image

libs/ui-react/src/lib/Components/MediaButton/MediaButton.stories.tsx
```diff
@@ -112,7 +112,7 @@ export const IconTypeShowcase: Story = {
           Flat icon (md)
         </MediaButton>
         <MediaButton
-          icon={<CryptoIcon ledgerId='bitcoin' ticker='BTC' size='32px' />}
+          icon={<CryptoIcon ledgerId='bitcoin' ticker='BTC' size={32} />}
```
aquelemiguel, line 115: you could potentially use your `cryptoIconSizes` instead of hardcoding

aquelemiguel (approved):
excellent work, not much to point out other than what copilot has commented

libs/ui-react/src/lib/Components/MediaButton/MediaButton.stories.tsx
aammami-ledger: this is the `MediaButton` not the `MediaImage`

PR #680, opened 2026-05-05
[D]DLS-702-feature(ui-react-visualization): Add Scrubber component

zel-kass:
## Overview

Introduce Scrubber component with keyboard controls, and plug it to CartesianChart.
Screenshot 2026-05-05 at 17 44 53


libs/ui-react-visualization/src/lib/Components/Axis/YAxis/YAxis.tsx
```diff
@@ -56,7 +56,10 @@ export function YAxis({
               y1={tick.position}
               x2={drawingArea.x + drawingArea.width}
               y2={tick.position}
-              style={{ stroke: cssVar('var(--border-muted-subtle)') }}
+              style={{
+                stroke: cssVar('var(--border-muted-subtle)'),
+                opacity: 0.5,
+              }}
```
aammami-ledger, line 61: so we do not have a token for this one ? looks like `border-muted-subtle` on figma, why do we need the reduce opacity ?

zel-kass: the exact token needed for this was border-muted-subtle-transparent, but when I tried to add it, the token wasn't recognized, so Laurine told me it was just border-muted-subtle with a 50% opacity

aquelemiguel (approved):
Regarding the gradient on the scrubber, I know we've brought this up in a daily before, but now I've had time to collect my ideas... I think the reason it looks odd is in this situation where the line intersects the lightest part:

Image

---

It looks way better here, where the intersect point is near the graph's equator so the gradient is in its most opaque part:

Image

--- 

I'd suggest confirming with Laurine if any of these solutions could work:
1. Remove gradient altogether, just use a solid line (we tried this before, mixed opinions).
2. Make it so the gradient always starts at the point of intersection, not in the equator of the graph (could be difficult dev-wise).
3. Remove both the gradient and the line (the change in transparency might be enough for the user to pinpoint the correct X coordinate).

Regardless, excellent work on this @zel-kass, as always. Nothing to point out code-wise, it's a bit cryptic and math heavy. Mostly tested interacting with it and feels great. 🫶

aammami-ledger (approved):
Beautiful work on this one 👏

PR #688, opened 2026-05-11
feat(ui-react): Add `ChartTooltip` to LineChart

aammami-ledger:


libs/ui-react-visualization/src/lib/Components/ChartTooltip/ChartTooltip.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+const annotatedIndices = new Set([4, 9]);
+
+const meta = {
+  component: ChartTooltip,
+  title: 'Visualization/ChartTooltip',
+  tags: ['experimental'],
+  decorators: [
+    (Story, context) => (
+      <StoryDecorator context={context}>
+        <div style={{ width: 600, padding: 16 }}>
```
zel-kass: [low] maybe just add a bg-base on this div so that we don't see the weird overlay effect on dark mode

libs/ui-react-visualization/src/lib/Components/ChartTooltip/ChartTooltip.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+      fireEvent.mouseLeave(svg);
+    });
+    expect(getByTestId('chart-tooltip').style.opacity).toBe('0');
+  });
+
+  it('positions the tooltip to the right of the scrubber by default', () => {
+    const { getByTestId } = renderTooltip({
+      tooltipProps: { items: () => [{ label: 'T', value: 'V' }] },
+    });
+    moveScrubber(getByTestId('chart-svg'), 0); // far left — no overflow
```
zel-kass: [low]: do we need the comments ?

libs/ui-react-visualization/src/lib/Components/ChartTooltip/ChartTooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+  BORDER_RADIUS,
+  DEFAULT_OFFSET,
+  DEFAULT_TOOLTIP_WIDTH,
+  PADDING_X,
+  PADDING_Y,
+  ROW_GAP,
+  ROW_HEIGHT,
+  TITLE_GAP,
+  TOOLTIP_TRANSITION,
+} from './constants';
```
zel-kass, line 17: [mid]: having a separate constants file is not a pattern we evolved with while implementing the charts, we usually colocate them at utils file top level. That said, I like it, good SoC, should we align the rest of the viz components to this ?

libs/ui-react-visualization/src/lib/Components/ChartTooltip/ChartTooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+        rx={BORDER_RADIUS}
+        fill={cssVar('var(--background-muted)')}
+      />
+      {hasTitle && (
+        <text
+          data-testid='chart-tooltip-title'
+          x={tooltipX + PADDING_X}
+          y={drawingArea.y + PADDING_Y + ROW_HEIGHT / 2}
+          dominantBaseline='middle'
+          style={{
```
zel-kass: [low]: for elegance purposes, we could extract this at module-level constants :
`const titleStyle = {
  fontSize: cssVar('var(--font-style-body-4-size)'),
  fontFamily: cssVar('var(--font-family-font)'),
  fill: cssVar('var(--text-base)'),
  fontWeight: cssVar('var(--font-style-body-4-weight-medium)'),
};`

maybe can do the same for the group style above but less shocking :
`const tooltipGroupStyle = {
  opacity: 1,
  transition: TOOLTIP_TRANSITION,
  pointerEvents: 'none' as const,
};`

libs/ui-react-visualization/src/lib/Components/ChartTooltip/ChartTooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+}: Readonly<ChartTooltipProps>) {
+  const { scrubberPosition } = useScrubberContext();
+  const { getXScale, getXAxisConfig, drawingArea } = useCartesianChartContext();
+
+  if (scrubberPosition === undefined) return HIDDEN_TOOLTIP;
+
+  const pixelX = resolvePixelX(scrubberPosition, getXScale, getXAxisConfig());
+  const resolvedItems: ChartTooltipItemData[] = items(scrubberPosition);
+
+  if (pixelX === undefined || resolvedItems.length === 0) return HIDDEN_TOOLTIP;
```
zel-kass: [mid]: we could avoid some computations by inverting the order here, checking if pixelX is undefined before calling the user provided function items().

libs/ui-react-visualization/src/lib/Components/ChartTooltip/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const DEFAULT_OFFSET = 10;
```
zel-kass, line 1: indeed this is clean af 🫨

zel-kass (approved):
Pre approvin, this is great work 🫶🏽

libs/ui-react-visualization/src/lib/Components/ChartTooltip/ChartTooltip.tsx
aammami-ledger: It’s just that I need to use the same constant across 2 files, and importing them from one file to another didn’t really make sense. That’s why I introduced it here  (also don’t have a dedicated utils file)

It's a pattern that already exists in the codebase, so IMHO yes i think we should align with it, when it makes sense ofc

PR #689, opened 2026-05-12
[D]DLS-700-feat(ui-react-visualization): Add ReferenceLine component

zel-kass:
## Overview

Introduce ReferenceLine component with vertical and horizontal configurations.

---

Screenshot 2026-05-12 at 12 05 45

### DIFF
--- a/libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
@@ -1,7 +1,6 @@
 import { cssVar } from '@ledgerhq/lumen-design-core';
 import { memo } from 'react';
 
-import { getPointOnScale } from '../../utils/scales/scales';
 import { useCartesianChartContext } from '../CartesianChart/context';
 
 import { DEFAULT_STROKE, DASH_ARRAY, STROKE_WIDTH } from './constants';
@@ -9,7 +8,7 @@ import type { ReferenceLineProps } from './types';
 import {
   computeHorizontalLabelCoordinates,
   computeVerticalLabelCoordinates,
-  isPixelWithinDrawingArea,
+  resolvePixel,
 } from './utils';
 
 const labelStyle = {
@@ -38,12 +37,8 @@ export const ReferenceLine = memo(
     const dashArray = lineStyle === 'dashed' ? DASH_ARRAY : undefined;
 
     if (dataY !== undefined) {
-      const yScale = getYScale();
-      if (!yScale) return null;
-
-      const yPixel = getPointOnScale(dataY, yScale);
-      if (!Number.isFinite(yPixel)) return null;
-      if (!isPixelWithinDrawingArea(yPixel, 'y', drawingArea)) return null;
+      const yPixel = resolvePixel(dataY, getYScale(), 'y', drawingArea);
+      if (yPixel === undefined) return null;
 
       const labelCoords = label
         ? computeHorizontalLabelCoordinates(
@@ -89,12 +84,8 @@ export const ReferenceLine = memo(
     }
 
     if (dataX !== undefined) {
-      const xScale = getXScale();
-      if (!xScale) return null;
-
-      const xPixel = getPointOnScale(dataX, xScale);
-      if (!Number.isFinite(xPixel)) return null;
-      if (!isPixelWithinDrawingArea(xPixel, 'x', drawingArea)) return null;
+      const xPixel = resolvePixel(dataX, getXScale(), 'x', drawingArea);
+      if (xPixel === undefined) return null;
 
       const labelCoords = label
         ? computeVerticalLabelCoordinates(
--- a/libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
@@ -18,8 +18,8 @@ const labelStyle = {
   fontFamily: cssVar('var(--font-family-font)'),
 };
 
-export const ReferenceLine = memo(
-  ({
+export const ReferenceLine = memo((props: Readonly<ReferenceLineProps>) => {
+  const {
     label,
     labelDx = 0,
     labelDy = 0,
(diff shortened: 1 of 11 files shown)


libs/ui-react-visualization/src/lib/Components/ReferenceLine/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+ * for a vertical reference line.
+ */
+export const computeVerticalLabelCoordinates = (
+  xPixel: number,
+  labelPosition: VerticalLabelPosition,
+  drawingArea: DrawingArea,
+  labelDx: number,
+  labelDy: number,
+  horizontalAlignment?: TextHorizontalAlignment,
+  verticalAlignment?: TextVerticalAlignment,
```
gamegee: [mid]: Clean-Code rules 
- -> prevent to have too many arguments (3 max)
- Better to have mandatory argument first, then optional arguments as object
- If all your arguments are mandatory, just create an object as it provide named keys that are easier to understand.

gamegee: [mid]: Also it seems the function is doing a lot:
- computing `x` & `y` 
- computing `textAnchor`
- computing `dominantBaseline`

The best course of action would be to split these functions so you have better separation of concerns and smaller functions

libs/ui-react-visualization/src/lib/Components/ReferenceLine/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+    default:
+      y = drawingArea.y + drawingArea.height;
+      defaultVerticalAlignment = 'bottom';
+      break;
+  }
+
+  return {
+    x: xPixel + labelDx,
+    y: y + labelDy,
+    textAnchor: resolveTextAnchor(horizontalAlignment ?? 'center'),
```
gamegee: [low]: 'center' is defined inline while for horizontal it refers to a variable.
Should we move both defaults to constants on top of the file ?

libs/ui-react-visualization/src/lib/Components/ReferenceLine/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+      defaultHorizontalAlignment = 'right';
+      break;
+  }
+
+  return {
+    x: x + labelDx,
+    y: yPixel + labelDy,
+    textAnchor: resolveTextAnchor(
+      horizontalAlignment ?? defaultHorizontalAlignment,
+    ),
```
gamegee: [low]: instead of a function, it could just be a mapping:

```tsx
const textAnchorMap = {
  left: 'start',
  right: 'end',
  center: 'middle'
}

textAnchor: textAnchorMap[horizontalAlignment ?? defaultHorizontalAlignment]
```

libs/ui-react-visualization/src/lib/Components/ReferenceLine/utils.ts
gamegee: Than being said, if you need to deduplicate function for horizontal & vertical, keep it as one is good

libs/ui-react-visualization/src/lib/Components/ReferenceLine/utils.ts
```diff
@@ -0,0 +1,157 @@
...
+): number | undefined => {
+  if (!scale) return undefined;
+
+  if (isCategoricalScale(scale) && scale(dataValue) === undefined)
+    return undefined;
+
+  const p = getPointOnScale(dataValue, scale);
+
+  if (!Number.isFinite(p)) return undefined;
+  if (!isPixelWithinDrawingArea(p, axis, drawingArea)) return undefined;
```
gamegee: [low]: group the "if":

```tsx
  if (
    !Number.isFinite(p) || 
    !isPixelWithinDrawingArea(p, axis, drawingArea)
  ) {
   return undefined;
  }
```

libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const { getXScale, getYScale, getXAxisConfig, getYAxisConfig, drawingArea } =
+    useCartesianChartContext();
+
+  const dashArray = lineStyle === 'dashed' ? DASH_ARRAY : undefined;
+
+  if (props.dataY !== undefined) {
+    const yValue = resolveDataValue(props.dataY, getYAxisConfig());
+    if (yValue === undefined) return null;
+    const yPixel = resolvePixel(yValue, getYScale(), 'y', drawingArea);
+    if (yPixel === undefined) return null;
```
gamegee, line 45: [mid]: lot of "if", what about having only one function here `resolveToPixel`
Also don't forget to pass all params as an object:

```tsx
// Not sure, you might rename few params

resolveToPixel({
  data: props.dataY,
  axisConfig: getYAxisConfig(),
  scale: getYScale,
  axis: 'y',
  drawingArea
})
```

libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
```diff
@@ -0,0 +1,140 @@
...
+          yPixel,
+          props.labelPosition ?? 'right',
+          drawingArea,
+          {
+            dx: labelDx,
+            dy: labelDy,
+            horizontalAlignment: labelHorizontalAlignment,
+            verticalAlignment: labelVerticalAlignment,
+          },
+        )
```
gamegee: [low]: same -> Define object named param, easier to read & understand for future maintainers

libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
```diff
@@ -0,0 +1,140 @@
...
+        )}
+      </g>
+    );
+  }
+
+  if (props.dataX !== undefined) {
+    const xValue = resolveDataValue(props.dataX, getXAxisConfig());
+    if (xValue === undefined) return null;
+    const xPixel = resolvePixel(xValue, getXScale(), 'x', drawingArea);
+    if (xPixel === undefined) return null;
```
gamegee, line 98: same

libs/ui-react-visualization/src/lib/Components/ReferenceLine/ReferenceLine.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    const labelCoords = label
+      ? computeVerticalLabelCoordinates(
```
gamegee: same

libs/ui-react-visualization/src/lib/Components/ReferenceLine/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  labelHorizontalAlignment?: TextHorizontalAlignment;
+  /**
+   * Controls how the label text anchors vertically relative to its position.
+   * - `'top'`: text hangs below the position
+   * - `'middle'`: text is vertically centered on the position
+   * - `'bottom'`: text sits above the position
+   *
+   * Defaults: `'bottom'` for horizontal lines, inferred from `labelPosition` for vertical lines.
+   */
+  labelVerticalAlignment?: TextVerticalAlignment;
```
gamegee: [mid]: 

## suggestion 1
Have `labelHorizontalAlignment` and `labelVerticalAlignment` included in the discrimination type, meaning that
- you can only define `labelHorizontalAlignment` when defining a vertical line 
- you can only define `labelVerticalAlignment` when defining a horizontal line 

(optional) -> Just have `start` `center` `end` on both type definition


## suggestion 2
- Remove `labelPosition` from the discrimination type and put it in the shared props
- The type def could be: `start` `center` `end` so it support RTL and also is shared by both version vertical and horizontal

gamegee (approved):

PR #694, opened 2026-05-13
[M]DLS-700-feat(ui-rnative-visualization): Add ReferenceLine component 

zel-kass:
## Overview

Introduce ReferenceLine component with vertical and horizontal configurations.

### DIFF
--- a/libs/ui-rnative-visualization/src/lib/Components/ReferenceLine/types.ts
+++ b/libs/ui-rnative-visualization/src/lib/Components/ReferenceLine/types.ts
@@ -58,8 +58,9 @@ export type VerticalLabelPosition = 'top' | 'middle' | 'bottom';
 
 export type HorizontalReferenceLineProps = ReferenceLineBaseProps & {
   /**
-   * Y-value in data space. The horizontal line is drawn at this value
-   * spanning the full width of the drawing area.
+   * Y-axis value in data space where the horizontal line is drawn.
+   * The horizontal line is rendered at this Y position spanning the full
+   * width of the drawing area.
    */
   dataY: number;
   /**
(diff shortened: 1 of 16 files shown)


libs/ui-rnative-visualization/src/lib/Components/ReferenceLine/types.ts
```diff
@@ -0,0 +1,84 @@
...
+   * - `'dashed'`: rounded dots with `strokeDasharray="0.1 6"`
+   * - `'solid'`: continuous line
+   * @default 'dashed'
+   */
+  lineStyle?: 'solid' | 'dashed';
+  /**
+   * Opacity applied to the line and label.
+   * @default 1
+   */
+  opacity?: number;
```
gamegee: [mid]: I didn't spotted that on react, but do we want to support opacity, while we already pass an hexadecimal to the stroke. Meaning we can use tokens with defined opacity as hexadecimal, e.g. `muted-transparent`

libs/ui-rnative-visualization/src/lib/Components/Scrubber/ScrubberProvider.tsx
```diff
@@ -87,9 +87,7 @@ export function ScrubberProvider({
         'worklet';
         isScrubbing.value = true;
         scheduleOnRN(handlePositionChange, e.x);
-      })
-      .onEnd(resetScrubber)
-      .onFinalize(resetScrubber);
```
gamegee, line 92: [high]:

Details
1. Removing onEnd/onFinalize from LongPress may leave scrubber stuck -- 9/10 Correctness -- ScrubberProvider.tsx:89-91

The PR removes .onEnd(resetScrubber) and .onFinalize(resetScrubber) from the LongPress gesture. This change is not mentioned in the PR description and appears unrelated to the ReferenceLine feature.

The risk: if a user long-presses (activating isScrubbing = true) but releases without panning, the Pan gesture is never manually activated, so its onEnd never fires. While Pan.onFinalize should fire when the gesture state machine resets, its behavior under manualActivation(true) when never activated is subtle and not guaranteed across all RNGH versions. Previously, LongPress.onEnd/onFinalize acted as a safety net for this scenario.

If the intent is to fix a double-reset race (both gestures calling resetScrubber), consider guarding resetScrubber to be idempotent instead:

const resetScrubber = useCallback((): void => {
  'worklet';
  if (!isScrubbing.value) return; // guard
  isScrubbing.value = false;
  scheduleOnRN(handlePositionChange, null);
}, [isScrubbing, handlePositionChange]);
Fix: Either restore onEnd/onFinalize on LongPress with an idempotent guard, or add a comment/test proving that Pan's onFinalize reliably fires in the no-pan case under manual activation.

gamegee (approved):

PR #697, opened 2026-05-15
feat(ui-rnative-visualization): add tooltip functionality

aammami-ledger:


apps/app-sandbox-rnative/src/app/blocks/LineCharts.tsx
```diff
@@ -374,6 +374,53 @@ export const LineCharts = () => {
...
+          />
+          <Point dataX={4} dataY={98} color='#47883A' label='ATH' />
+          <Point
+            dataX={9}
+            dataY={4}
+            color='#C24244'
+            label='Low'
+            labelPosition='bottom'
+          />
+        </LineChart>
```
gamegee: [mid]:
You could define the points array once and derive both the Point components and the Scrubber tooltip logic from it. Something like:
```tsx
const points = [
  { dataX: 4, dataY: 98, color: '#47883A', label: 'ATH' },
  { dataX: 9, dataY: 4, color: '#C24244', label: 'Low', labelPosition: 'bottom' as const },
];
Then use it in both places:

<Scrubber
  tooltip={(i) => {
    const point = points.find((p) => p.dataX === i);
    if (!point) return { items: [] };
    return {
      items: [
        { label: 'Month', value: months[i] },
        { label: 'Price', value: `$${sampleSeries[0].data[i]}` },
      ],
    };
  }}
/>
{points.map((point) => (
  <Point key={point.dataX} {...point} />
))}
```

This removes the hardcoded index checks (i !== 4 && i !== 9) and the duplicated point definitions. Adding or removing a point becomes a single change to the points array.

gamegee: Same for react if already defined like this

libs/ui-rnative-visualization/src/lib/Components/Scrubber/types.ts
```diff
@@ -35,6 +37,112 @@ export type ScrubberProviderProps = {
...
+  /**
+   * Bounding box of the chart's drawing area (x, y, width, height). Used to
+   * constrain the tooltip so it never overflows the chart boundaries.
+   */
+  drawingArea: DrawingArea;
+  /**
+   * Zero-based index of the data point currently under the scrubber. Maps
+   * directly to the series data arrays so consumers can look up values.
+   */
+  dataIndex: number;
```
gamegee: [mid] Seems that you don't need to pass the `dataIndex` to the tooltip as its not used.
Maybe to confirm also on react

libs/ui-rnative-visualization/src/lib/Components/Scrubber/types.ts
```diff
@@ -35,6 +37,112 @@ export type ScrubberProviderProps = {
...
+  /**
+   * Optional title displayed at the top of the tooltip. Pass `null` or omit
+   * to render the tooltip without a title row.
+   */
```
gamegee: [low]: I would only allow `string | number | undefined` -> `undefined is already implicit with `title?:` question mark in ts

libs/ui-rnative-visualization/src/lib/Components/Scrubber/types.ts
```diff
@@ -35,6 +37,112 @@ export type ScrubberProviderProps = {
...
+ */
+export type ScrubberTooltipContent = ScrubberTooltipLayoutProps & {
+  /**
+   * Optional header. Static value or callback per data index.
+   * A callback may return `null` or `undefined` to suppress the title.
+   */
+  title?:
+    | string
+    | number
+    | ((index: number) => string | number | null | undefined);
```
gamegee: [low]: same remove `null`

libs/ui-rnative-visualization/src/lib/Components/Scrubber/types.ts
gamegee: By doing that, you can just remove condition checking the `null` in implementation.

---
Note: double check on react

apps/app-sandbox-rnative/src/app/blocks/LineCharts.tsx
```diff
@@ -430,6 +432,65 @@ const ReferenceLineBasic = () => {
...
+        }}
+      />
+      {annotatedPoints.map((point) => (
+        <Point key={point.dataX} {...point} />
+      ))}
+    </LineChart>
+  </Section>
+);
+
+const LineChartTooltipAlwaysVisible = () => {
```
aquelemiguel: could be performance-related because of how many graphs we're displaying on this sandbox page, but reporting on the off-chance it is some rendering bug:

https://github.com/user-attachments/assets/1f193616-4962-4d65-822e-324020a16456

https://github.com/user-attachments/assets/69044b1a-8583-4682-a897-1d8d6e4324e1

some layout issues appear randomly as seen above, including crashing. was unable to get it on a recording but i also got instances where the price text was overflowing out of the tooltip width

gamegee (approved):

PR #698, opened 2026-05-15
feat(ui-rnative): add search into OptionList

aquelemiguel:
Closes [DLS-712](https://ledgerhq.atlassian.net/browse/DLS-712).

## Demo (iOS simulator)
iPhone 17, iOS 26.5

https://github.com/user-attachments/assets/2dfbcb90-6001-47b2-bc85-c866990f277d

[DLS-712]: https://ledgerhq.atlassian.net/browse/DLS-712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-rnative/src/lib/Components/OptionList/OptionList.mdx
+++ b/libs/ui-rnative/src/lib/Components/OptionList/OptionList.mdx
@@ -74,6 +74,14 @@ Search works within grouped items too. Empty groups are hidden automatically:
 
 <Canvas of={OptionListStories.WithSearchAndGroups} />
 
+Pass a custom `filter` to match against fields beyond the label. Here, the query also matches each item's ticker via `meta`:
+
+<Canvas of={OptionListStories.WithCustomSearchFilter} />
+
+Drive the search input from the outside with `searchValue` + `onSearchValueChange`. This is useful for syncing search state with other UI or kicking off async fetches:
+
+<Canvas of={OptionListStories.WithControlledSearch} />
+
 ### Trigger showcase
 
 OptionList can be opened from any trigger. `MediaButton` supports multiple appearances (`gray`, `transparent`, `no-background`), optional icons (`flat` / `rounded`), and disabled state:
--- a/libs/ui-rnative/src/lib/Components/OptionList/OptionList.mdx
+++ b/libs/ui-rnative/src/lib/Components/OptionList/OptionList.mdx
@@ -36,6 +36,7 @@ It handles **selection state**, **automatic grouping** (via a `group` field on i
 - **OptionListItemContentRow**: Horizontal row for placing elements side-by-side (e.g. title + tag)
 - **OptionListItemText**: Styled title text
 - **OptionListItemDescription**: Styled description text
+- **OptionListSearch**: Search input that filters items by label (default) or via a custom `filter` function
 
 ## Properties
 
@@ -62,6 +63,16 @@ Individual items can be disabled:
 Combining grouping with rich item content:
 
 <Canvas of={OptionListStories.GroupedWithContentRow} />
+
+### Search
+
+Drop `OptionListSearch` inside the list to enable case-insensitive filtering on item labels. Pair with `OptionListEmptyState` to handle no-match scenarios. Pass a custom `filter` to match against meta (or anything else), or `filter={null}` to disable filtering. Use `filteredItems` + `onSearchValueChange` for async/server-side search.
+
+<Canvas of={OptionListStories.WithSearch} />
+
+Search works within grouped items too. Empty groups are hidden automatically:
+
+<Canvas of={OptionListStories.WithSearchAndGroups} />
 
 ### Trigger showcase
 
--- a/libs/ui-rnative/src/lib/Components/OptionList/types.ts
+++ b/libs/ui-rnative/src/lib/Components/OptionList/types.ts
@@ -39,7 +39,6 @@ export type OptionListContextValue = {
   groups: OptionListItemGroup[];
   flatItems: OptionListItemData[];
   resolvedSearchValue: string;
-  registerSearch: () => () => void;
   handleSearchValueChange: (value: string) => void;
 };
 
--- a/libs/ui-rnative/src/lib/Components/OptionList/types.ts
+++ b/libs/ui-rnative/src/lib/Components/OptionList/types.ts
@@ -4,6 +4,7 @@ import type {
   StyledTextProps,
   StyledViewProps,
 } from '../../../styles';
+import type { SearchInputProps } from '../SearchInput';
 
 export type MetaShape = Record<string, unknown>;
 
@@ -37,6 +38,9 @@ export type OptionListContextValue = {
   isGrouped: boolean;
   groups: OptionListItemGroup[];
   flatItems: OptionListItemData[];
+  resolvedSearchValue: string;
+  registerSearch: () => () => void;
+  handleSearchValueChange: (value: string) => void;
 };
 
 /** Internal type -- consumers never construct this directly. */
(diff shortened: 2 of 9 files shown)


libs/ui-rnative/src/lib/Components/OptionList/types.ts
```diff
@@ -103,6 +139,11 @@ export type OptionListEmptyStateProps = {
   description?: string;
 } & Omit<StyledViewProps, 'children'>;
 
+export type OptionListSearchProps = Omit<
+  SearchInputProps,
+  'value' | 'onChangeText' | 'defaultValue'
+>;
```
aammami-ledger, line 145: better to Pick rather then Omit, because if we add new props to SearchInputProps they will be added here and we might not want that. 

```suggestion
export type OptionListSearchProps = Pick<
  SearchInputProps,
  | 'placeholder'
  | 'className'
  | 'helperText'
  | 'status'
  | 'aria-invalid'
  | 'suffix'
  | 'onClear'
  | 'hideClearButton'
>;
```

libs/ui-rnative/src/lib/Components/OptionList/OptionList.mdx
```diff
@@ -63,6 +64,16 @@ Combining grouping with rich item content:
 
 <Canvas of={OptionListStories.GroupedWithContentRow} />
 
+### Search
```
aammami-ledger, line 67: What about we add more exemples: one with custom filter and one controlled search ?

libs/ui-rnative/src/lib/Components/OptionList/types.ts
aquelemiguel: i see your point, however i'm not keen on exhaustive listing of all basic props, e.g., focus handlers like `onBlur`, `onFocus` or other core functionality like `testID`, `accessibilityLabel`, etc.
is SearchInputProps not stable? i'd imagine the pick list would go stale fast

libs/ui-rnative/src/lib/Components/OptionList/OptionList.mdx
aquelemiguel: good ones, i've added these in https://github.com/LedgerHQ/lumen/pull/698/commits/29632a7bbb24eb9496e3e207ff681657294f94e1

libs/ui-rnative/src/lib/Components/OptionList/types.ts
aammami-ledger: it's stable for now, but it was one of the first components we build I think it's likely we will introduce some changes to it. 

But I see what you mean, well Pick is what I did on React, maybe we can inline them.

aammami-ledger (approved):
Approving this! I would wait for a second review before merging though 👍

libs/ui-rnative/src/lib/Components/OptionList/types.ts
gamegee: If we go for `Pick<T>` it should be followed by an extension of the core HTML attributes, so at the end:

```tsx
type OptionListSearchProps = Pick<SearchInputProps, '...' | '...'> 
  & ComponentPropsWithRef<'input'>;
```

gamegee: It depends on the use-case, because something you might want to only Omit one property, so that the component that is "built-on-top-of" will benefit new API changes of the component automatically. 

For some other component "built-on-top-of" we might prefer `Pick` because we know we extends only a narrowed API of the base component.

TLDR;
It depends, the rule-of-thumb is not easy to define, we should always go for the solution that ease the long-term-maintenance and prevent us to look at multiple components when adding/updating a prop

Following that, for our current use-case I believe `Omit` could work

libs/ui-rnative/src/lib/Components/OptionList/useOptionList/useOptionListItems.ts
```diff
@@ -24,26 +25,68 @@ const groupByField = <TMeta extends MetaShape = MetaShape>(
...
-): UseOptionListItemsResult<TMeta> => {
-  const isGrouped = hasGroups(items);
-  return isGrouped
-    ? { isGrouped: true, groups: groupByField(items), flatItems: [] }
-    : { isGrouped: false, groups: [], flatItems: items };
-};
+export const defaultLabelFilter = (
+  item: OptionListItemData,
+  query: string,
+): boolean => item.label.toLowerCase().includes(query.toLowerCase());
```
gamegee, line 31: [mid]: after investigating how we could provide a bit more optimized search algorithm, Claude ends up with that.

## Simple solution 

```tsx
const searchOnce = (items, query, getLabel = (i) => i.label) => {
  const q = normalize(query ?? "");
  if (!q) return items;
  const tokens = q.split(/\s+/).filter(Boolean);
  return items.filter((item) => {
    const haystack = normalize(getLabel(item) ?? "");
    return tokens.every((t) => haystack.includes(t));
  });
};
```

## Advanced solution with index. 

This solution includes a buildIndex by label, which is run only when defining the data at load.


```tsx
/**
 * Normalize a string for search: lowercase, trim, remove diacritics.
 */
const normalize = (str) =>
  str
    .toLowerCase()
    .normalize("NFD")
    .replace(/\p{Diacritic}/gu, "")
    .trim();

/**
 * Build a searchable index from a list of items.
 * Call this once when the data loads (or when it changes).
 *
 * @param {Array} items - Source items
 * @param {Function} getLabel - Extracts the searchable string from an item
 *                              (defaults to item.label)
 * @returns {Array} Items augmented with a `_searchable` field
 */
const buildIndex = (items, getLabel = (item) => item.label) =>
  items.map((item) => ({
    item,
    _searchable: normalize(getLabel(item) ?? ""),
  }));

/**
 * Filter an indexed list against a query.
 * Supports multi-word queries: every token must match (AND logic).
 *
 * @param {Array} indexed - Output of buildIndex()
 * @param {string} query - User input
 * @returns {Array} Matching original items
 */
const search = (indexed, query) => {
  const q = normalize(query ?? "");
  if (!q) return indexed.map(({ item }) => item);

  const tokens = q.split(/\s+/).filter(Boolean);
  return indexed
    .filter(({ _searchable }) => tokens.every((t) => _searchable.includes(t)))
    .map(({ item }) => item);
};
```

libs/ui-rnative/src/lib/Components/OptionList/types.ts
```diff
@@ -46,16 +49,49 @@ export type OptionListItemGroup<TMeta extends MetaShape = MetaShape> = {
...
+  /**
+   * Custom item/query matcher.
+   * Defaults to case-insensitive label match; `null` disables filtering.
+   */
+  filter?: null | ((item: OptionListItemData<TMeta>, query: string) => boolean);
+  /**
+   * Pre-filtered items for async/remote search.
+   * Bypasses internal filtering.
+   */
+  filteredItems?: OptionListItemData<TMeta>[];
```
gamegee, line 82: Question: Does this props allow to have manual filter, meaning they will only plug `value` and `onChange` from the search and make the filter on back-end side instead of client-side

aquelemiguel: it does! you can pre-filter on the backend and pass it to `filteredItems`, you'd use `searchValue` + `onSearchValueChange` to trigger a backend refetch

libs/ui-rnative/src/lib/Components/OptionList/useOptionList/useOptionListItems.ts
aquelemiguel: IMO i think currently it does a nice job as a default, just a simple predictable case-insensitive substring match, plus we already have an escape hatch via `filter` prop if consumers want to tweak the algorithm

libs/ui-rnative/src/lib/Components/OptionList/types.ts
gamegee: Oh ok nice then 👍 
If its clear on the props + the story how to do it then I am good

libs/ui-rnative/src/lib/Components/OptionList/useOptionList/useOptionListItems.ts
gamegee: Yes I was thinking of already proposing an optimized algorithm as strong default.
The normalize + early returns could help

PR #701, opened 2026-05-19
fix(ui-rnative): fix header overlap with absolutely positioned gradient inside BottomSheet

aquelemiguel:
Closes [DLS-711](https://ledgerhq.atlassian.net/browse/DLS-711).

Proposes exposing Gorhom's `backgroundComponent` prop to take in e.g., a gradient.
See https://gorhom.dev/react-native-bottom-sheet/custom-background.

Also adds a prop to hide the custom handle, replacing it with an invisible one to not cause layout shift.

```tsx
<BottomSheet
  {...props}
  { /** ... */ }
  backgroundComponent={<LinearGradient />} // <-- new prop
  hideHandle // <-- new prop
>
  <BottomSheetView>
    <BottomSheetHeader { /** ... */ } />
    <Box lx={{ padding: 's16' }}>
      <Text typography='body2' lx={{ color: 'base' }}>
        { /** ... */ }
      </Text>
    </Box>
  </BottomSheetView>
</BottomSheet>
```

## Also

* After syncing with Laurine, updated CustomHandle color to `mutedTransparentPressed` so it provides good contrast regardless of the background content.
* Took the liberty to migrate from `ElementRef` to `ComponentRef` following some deprecation warnings.

## Demo (iOS sandbox)

image

## Considerations

This is the issue originally reported by Lucas. Absolutely positioning a node inside of the bottom sheet children results on the element being hidden behind the BottomSheetHeader.

Screenshot 2026-05-19 at 16 26 11

---

Also note that regardless of making the header's background transparent, the 16px row wrapping the CustomHandle takes precedence in rendering.

Screenshot 2026-05-19 at 17 44 43

---

Thus, there's no way of achieving [this intended look](https://www.figma.com/design/NJZI8sW7CTSZWJcZ95zEEj/Core-Device-Actions-%E2%80%A2-Wallet?node-id=7269-3701&m=dev) without exposing Gorhom's BottomSheet `backgroundComponent` prop.

image

[DLS-711]: https://ledgerhq.atlassian.net/browse/DLS-711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-rnative/src/lib/Components/BottomSheet/BottomSheet.tsx
+++ b/libs/ui-rnative/src/lib/Components/BottomSheet/BottomSheet.tsx
@@ -41,9 +41,6 @@ const useStyles = ({
           flex: 1,
           borderTopLeftRadius: t.borderRadius.xl,
           borderTopRightRadius: t.borderRadius.xl,
-          backgroundColor: hasCustomBackground
-            ? 'transparent'
-            : t.colors.bg.canvasSheet,
           overflow: 'hidden',
         },
         shadow && {
--- a/libs/ui-rnative/src/lib/Components/BottomSheet/BottomSheet.tsx
+++ b/libs/ui-rnative/src/lib/Components/BottomSheet/BottomSheet.tsx
@@ -25,7 +25,13 @@ const MAX_DYNAMIC_CONTENT_SIZE = {
   fullWithOffset: FULL_WITH_OFFSET,
 };
 
-const useStyles = ({ shadow }: { shadow: boolean }) => {
+const useStyles = ({
+  shadow,
+  hasCustomBackground,
+}: {
+  shadow: boolean;
+  hasCustomBackground: boolean;
+}) => {
   return useStyleSheet(
     (t) => ({
       root: StyleSheet.flatten([
@@ -35,7 +41,10 @@ const useStyles = ({ shadow }: { shadow: boolean }) => {
           flex: 1,
           borderTopLeftRadius: t.borderRadius.xl,
           borderTopRightRadius: t.borderRadius.xl,
-          backgroundColor: t.colors.bg.canvasSheet,
+          backgroundColor: hasCustomBackground
+            ? 'transparent'
+            : t.colors.bg.canvasSheet,
+          overflow: 'hidden',
         },
         shadow && {
           boxShadow: t.shadows.lg,
@@ -46,7 +55,7 @@ const useStyles = ({ shadow }: { shadow: boolean }) => {
         backgroundColor: t.colors.bg.canvasSheet,
       },
     }),
-    [shadow],
+    [shadow, hasCustomBackground],
   );
 };
 
@@ -74,6 +83,7 @@ export const BottomSheet = ({
   onBackdropPress,
   onChange,
   snapPoints = 'fullWithOffset',
+  backgroundComponent,
   ref,
   ...props
 }: BottomSheetProps) => {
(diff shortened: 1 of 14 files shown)


libs/ui-rnative/src/lib/Components/BottomSheet/types.ts
```diff
@@ -131,8 +138,22 @@ export type BottomSheetProps = PropsWithChildren & {
...
+   * Custom background component rendered behind the sheet content and handle.
+   * Use to render gradients or other effects that should span the full sheet area,
+   * including the handle strip. When provided, the default sheet background is removed.
+   * @default undefined
+   */
+  backgroundComponent?: FC<BottomSheetBackgroundProps> | null;
+  /**
+   * If true, the drag handle (grabber) at the top of the sheet is hidden.
+   * @default false
+   */
```
gamegee, line 151: [high]: is it validated by the design ? It seems weird to have a bottom-sheet without it.
And it seems your background works with the handle, so we could just keep it

libs/ui-rnative/src/lib/Components/BottomSheet/types.ts
```diff
@@ -5,16 +5,23 @@ import type {
...
+  Ref,
+} from 'react';
+
 import type { StyledViewProps } from '../../../styles';
 export type BottomSheetProps = PropsWithChildren & {
   /**
    * Ref to the bottom sheet component.
    */
-  ref?: Ref<React.ElementRef<typeof GorhomBottomSheetModal>>;
+  ref?: Ref<ComponentRef<typeof GorhomBottomSheetModal>>;
```
gamegee, line 24: 👍

libs/ui-rnative/src/lib/Components/BottomSheet/CustomHandle.tsx
```diff
@@ -28,12 +28,33 @@ const useStyles = () => {
...
+
+  return (
+    <View
+      {...props}
+      ref={ref}
+      style={styles.container}
+      testID='bottom-sheet-handle-hidden'
+    />
+  );
+};
```
gamegee, line 60: Question: Can't we just return null instead of stylizing a view ?

libs/ui-rnative/src/lib/Components/BottomSheet/BottomSheet.tsx
```diff
@@ -35,7 +41,10 @@ const useStyles = ({ shadow }: { shadow: boolean }) => {
           flex: 1,
           borderTopLeftRadius: t.borderRadius.xl,
           borderTopRightRadius: t.borderRadius.xl,
-          backgroundColor: t.colors.bg.canvasSheet,
+          backgroundColor: hasCustomBackground
+            ? 'transparent'
+            : t.colors.bg.canvasSheet,
```
gamegee: [mid]: what about using it always transparent, it should use the one from the parent ?

libs/ui-rnative/src/lib/Components/BottomSheet/types.ts
aquelemiguel: spoke with laurine and she has validated but we may bring it up in tomorrow's daily
perhaps as long as it includes the close button it's fair game UX wise

libs/ui-rnative/src/lib/Components/BottomSheet/CustomHandle.tsx
aquelemiguel: the muted top bar seen below only exists because of the handle. if we returned null, then the sheet's content would be shifted up which i felt like an unintended side-effect, maybe?

image

libs/ui-rnative/src/lib/Components/BottomSheet/BottomSheet.tsx
aquelemiguel: nice catch, from my testing, it doesn't even need this style

gamegee (approved):

PR #702, opened 2026-05-20
refactor(ui-react): replace Radix UI DropdownMenu with Base UI Menu

aammami-ledger:


package.json
```diff
@@ -44,7 +44,6 @@
     "@nx/workspace": "22.5.3",
     "@radix-ui/react-checkbox": "^1.3.2",
     "@radix-ui/react-dialog": "^1.1.15",
-    "@radix-ui/react-dropdown-menu": "^2.1.2",
```
gamegee, line 47: 🫶

gamegee (commented):
Seems good, I ran the chromatic on your PR

Also, maybe we could have a story where the menu is already open - so chromatic could  could catch inner menu UI regression. I am thinking of the **complete example** story maybe ?

---
Also lacking the NX version plan with **migration guide**

gamegee (approved):

PR #703, opened 2026-05-20
[D]DLS-728-feat(ui-react-visualisation): Add point magnetisation

zel-kass:
## Overview

Introduce point magnetisation logic, with a default radius of 8px.

### DIFF
--- a/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
+++ b/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
@@ -69,7 +69,7 @@ export type CartesianChartProps = {
    * Pixel radius within which the scrubber magnetically snaps to registered
    * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
    * Set to `0` to disable magnetisation.
-   * @default 10
+   * @default 6
    */
   magnetRadius?: number;
 };
--- a/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
+++ b/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
@@ -68,7 +68,7 @@ export type CartesianChartProps = {
   /**
    * Pixel radius within which the scrubber magnetically snaps to registered
    * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
-   * Set to `0` to disable magnetisation.
+   * Set to `0` to disable magnetization.
    * @default 8
    */
   magnetRadius?: number;
--- a/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
+++ b/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
@@ -69,7 +69,7 @@ export type CartesianChartProps = {
    * Pixel radius within which the scrubber magnetically snaps to registered
    * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
    * Set to `0` to disable magnetisation.
-   * @default 6
+   * @default 8
    */
   magnetRadius?: number;
 };
--- a/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
+++ b/libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
@@ -65,4 +65,11 @@ export type CartesianChartProps = {
    * @default true
    */
   animate?: boolean;
+  /**
+   * Pixel radius within which the scrubber magnetically snaps to registered
+   * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
+   * Set to `0` to disable magnetisation.
+   * @default 10
+   */
+  magnetRadius?: number;
 };
--- a/libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
+++ b/libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
@@ -81,7 +81,7 @@ export type LineChartProps = {
    * Pixel radius within which the scrubber magnetically snaps to registered
    * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
    * Set to `0` to disable magnetisation.
-   * @default 10
+   * @default 6
    */
   magnetRadius?: number;
 };
--- a/libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
+++ b/libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
@@ -77,4 +77,11 @@ export type LineChartProps = {
    * @default true
    */
   animate?: boolean;
+  /**
+   * Pixel radius within which the scrubber magnetically snaps to registered
+   * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
+   * Set to `0` to disable magnetisation.
+   * @default 10
+   */
+  magnetRadius?: number;
 };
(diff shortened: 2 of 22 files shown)


libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
```diff
@@ -77,4 +77,11 @@ export type LineChartProps = {
    * @default true
    */
   animate?: boolean;
+  /**
+   * Pixel radius within which the scrubber magnetically snaps to registered
+   * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
+   * Set to `0` to disable magnetisation.
+   * @default 6
+   */
+  magnetRadius?: number;
```
gamegee, line 86: I would challenge with Laurine to go for 8 - wdyt ?

libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
```diff
@@ -65,4 +65,11 @@ export type CartesianChartProps = {
    * @default true
    */
   animate?: boolean;
+  /**
+   * Pixel radius within which the scrubber magnetically snaps to registered
+   * magnetic `<Point>` components. Requires `enableScrubbing` to be `true`.
+   * Set to `0` to disable magnetisation.
+   * @default 6
+   */
+  magnetRadius?: number;
```
gamegee, line 74: [mid]: The term `radius` might be improved, thinking of `magnetThreshold` or `magnetDistance` for accuracy

libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
zel-kass: Hagrid

gamegee: [mid]: the `magnetRadius` is also duplicated on Scrubber, meaning you can have a mismatch

```tsx
<LineChart magnetRadius={6}>
  <Scrubber magnetRadius={100} />
</LineChart>
```

Also it make it unclear which one will impact the scrubber magnetism with the Points

libs/ui-react-visualization/src/lib/Components/Scrubber/types.ts
```diff
@@ -31,6 +31,12 @@ export type ScrubberProviderProps = {
    * Optional external callback fired whenever the scrubber position changes.
    */
   onScrubberPositionChange?: (index: number | undefined) => void;
+  /**
+   * Pixel radius within which the scrubber magnetically snaps to registered
+   * magnetic points. Set to `0` to disable magnetisation.
+   * @default 6
+   */
+  magnetRadius?: number;
```
gamegee, line 39: see comment on the lineChart types

libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
gamegee: nvm, missread the Scrubber & SrubberProvider, its just props-drill, so all fine

gamegee (commented):
# Feedbacks
Additional suggestions


## Type assertion
`Point/utils.ts:54`

const index = (axisData as number[]).indexOf(dataX);
AxisConfigProps.data is typed string[] | number[]. If a band scale uses string categories (e.g., dates as ISO strings), this cast silently breaks — indexOf would compare number against string and always return -1. The function won't crash, but magnetisation would silently fail.

Fix: Add a runtime guard:

```tsx 
if (!axisData || typeof axisData[0] !== 'number') return dataX;
const index = (axisData as number[]).indexOf(dataX);
```
Or use a type predicate helper that already exists in the codebase (isNumberArray in Scrubber/utils.ts).

gamegee (commented):
Missing nx plan

libs/ui-react-visualization/src/lib/Components/Scrubber/utils.ts
```diff
@@ -121,3 +121,33 @@ export const resolvePixelX = (
...
+
+  for (const magnetIndex of magneticIndices) {
+    const magnetPixelX = getPixelForIndex(magnetIndex);
+    if (magnetPixelX === undefined) continue;
+    const distance = Math.abs(pixelX - magnetPixelX);
+    if (distance <= magnetRadius && distance < closestDistance) {
+      closestDistance = distance;
+      closestMagneticIndex = magnetIndex;
+    }
+  }
```
gamegee: [mid/high]: There is room for improvement in the performances.

You've resolved the invocation issue with rAF - but you can also apply additional improvements.

## Precompute + Sort
As discussed the other day, one improvement would be to precompute magnet pixel position with that simple interface:

```tsx
type MagnetEntry = { index: number; pixelX: number };
```

Magnet pixel positions only change when the xScale / dimensions / magneticIndices / axis data change, not on every cursor move. Lift that work into a useMemo so each pointer event becomes a pure lookup.

Could look like that:
```tsx
/**
 * Builds a pixel-sorted view of the magnetic indices.
 * Runs once per scale / dimensions / magnet-set change — never on pointer move.
 */
export const buildSortedMagnets = (
  magneticIndices: ReadonlySet<number>,
  getPixelForIndex: (index: number) => number | undefined,
): MagnetEntry[] => {
  if (magneticIndices.size === 0) return [];
  const entries: MagnetEntry[] = [];
  for (const index of magneticIndices) {
    const pixelX = getPixelForIndex(index);
    if (pixelX !== undefined) entries.push({ index, pixelX });
  }
  entries.sort((a, b) => a.pixelX - b.pixelX);
  return entries;
};
```
APPLY
```tsx
/**
 * Given a resolved data index and the raw pixel position, checks whether any
 * magnetic point is within `magnetRadius` pixels of `pixelX`. If so, returns
 * the closest magnetic index; otherwise returns the original `resolvedIndex`.
 *
 * Pre-condition: `sortedMagnets` is sorted ascending by `pixelX`.
 */
export const applyMagnetisation = (
  resolvedIndex: number,
  pixelX: number,
  sortedMagnets: ReadonlyArray<MagnetEntry>,
  magnetRadius: number,
): number => {
  if (magnetRadius <= 0 || sortedMagnets.length === 0) return resolvedIndex;
  let closestMagneticIndex = resolvedIndex;
  let closestDistance = Infinity;
  for (const magnet of sortedMagnets) {
    const distance = magnet.pixelX - pixelX;
    if (distance > magnetRadius) break;
    if (distance < -magnetRadius) continue;
    const absDistance = distance < 0 ? -distance : distance;
    if (absDistance < closestDistance) {
      closestDistance = absDistance;
      closestMagneticIndex = magnet.index;
    }
  }
  return closestMagneticIndex;
};
```

Precompute-memoization
```tsx
const magnetKey = useMemo(
  () => Array.from(magneticIndices).sort((a, b) => a - b).join(','),
  [magneticIndices],
);

const sortedMagnets = useMemo(
  () =>
    buildSortedMagnets(magneticIndices, (i) =>
      resolvePixelX(i, getXScale, getXAxisConfig()),
    ),
  [magnetKey, getXScale, getXAxisConfig],
);

```

---
## Additional improvements
**One other interesting improvement is to decouple visual from react states.** (I've learn about this one)

If scrubberPosition is in React state, every pointer move re-renders the chart, tooltip, beacon, overlay, and any subscriber to the context. For a line chart with 1k+ points that's the real cost - not the magnet search.

Two common patterns:

- Imperative ref-driven layer: keep scrubberPosition in a ref + emit via a tiny event bus / useSyncExternalStore; the chart subscribes selectively. Only components that visually need the position re-render.
- CSS transform / direct DOM write on a single SVG <g transform="translate(x,0)"> for the beacon + crosshair, and only setState on drag end / discrete tooltip changes.
Even just splitting the context into "scrubbing pixel" (high-frequency, ref) and "scrubber index" (low-frequency, state) goes a long way.


---
We can dig together tomorrow 👍 

Especially on the 2nd pattern, I am not super familiar with the solutions provided - but I've already had this kind of usecases.

gamegee (approved):

PR #709, opened 2026-05-22
[M]DLS-728-feat(ui-rnative-visualisation): Add point magnetisation

zel-kass:
## Overview

Introduce point magnetisation logic, with a default radius of 6px.


libs/ui-rnative-visualization/src/lib/Components/Scrubber/ScrubberProvider.tsx
```diff
@@ -78,15 +96,28 @@ export function ScrubberProvider({
...
+          pixelX,
+          sortedMagnets,
+          magnetRadius,
+        });
+      }
+
       setScrubberPositionAndNotify(index);
     },
-    [setScrubberPositionAndNotify],
+    [magnetRadius, setScrubberPositionAndNotify, sortedMagnets],
```
gamegee: [mid]: I would not have found this one from Opus 4.8
⏬ 


`sortedMagnets` in handlePositionChange deps breaks gesture stability pattern — 7/10 Performance - `ScrubberProvider.tsx:96`

The existing ScrubberProvider was carefully designed to keep handlePositionChange reference-stable by reading volatile values through latest.current. The PR adds sortedMagnets and magnetRadius directly to the useCallback dependency array:

```tsx
ScrubberProvider.tsx
Lines 95-105
    const ref = latest.current;
    const scale = ref.getXScale();
    if (!scale || ref.dataLength <= 0) return;
    // ...
    if (magnetRadius > 0) {
      index = applyMagnetisation({
        resolvedIndex: index,
        pixelX,
        sortedMagnets,
        magnetRadius,
      });
```  

Since sortedMagnets is recomputed on every version bump (each Point mount), this triggers the cascade: `handlePositionChange → resetScrubber → composed → surface` — tearing down and rebuilding the GestureDetector for each magnetic point registration. With 6 magnetic points (as in the sandbox example), that's 6 full gesture teardowns during mount.


Fix: Store sortedMagnets and magnetRadius in latest.current so handlePositionChange stays reference-stable:
```tsx
latest.current.sortedMagnets = sortedMagnets;
latest.current.magnetRadius = magnetRadius;
```

Then read from ref.sortedMagnets / ref.magnetRadius inside the callback, removing them from the dep array.

gamegee (approved):
✅

PR #712, opened 2026-05-25
feat(ui-react/ui-rnative): support disabled state for MediaImage + DotSymbol + DotIcon

aquelemiguel:
Closes [DLS-734](https://ledgerhq.atlassian.net/browse/DLS-734).

## Considerations

* Regarding the `<DisabledProvider value={{ disabled: false }}>` pattern, the goal is to apply the opacity to the outer wrapper and reset for the children.
If you allow it to propagate, each child applies it own opacity, resulting in this:

image

## Demo (iOS sandbox)

Showcasing disabled provider working correctly, provided e.g., by ListItem and picked up by MediaImage in isolation and wrapped inside DotSymbol.

Tested in iPhone 17 simulator running iOS 26.5.

https://github.com/user-attachments/assets/245e1c43-c2de-4ab2-bb64-e2add653fd79

[DLS-734]: https://ledgerhq.atlassian.net/browse/DLS-734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
+++ b/libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
@@ -32,10 +32,6 @@ const dotVariants = cva(
         muted: 'bg-muted-strong',
         error: 'bg-error-strong',
       },
-      disabled: {
-        true: 'opacity-30',
-        false: '',
-      },
     },
     compoundVariants: [
       { size: 16, shape: 'square', className: 'rounded-[5px]' },
@@ -111,17 +107,19 @@ export const DotIcon = ({
   });
 
   return (
-    <DisabledProvider value={{ disabled }}>
+    <DisabledProvider value={{ disabled: false }}>
       <div
         ref={ref}
-        className={cn('relative inline-flex w-fit', className)}
+        className={cn(
+          'relative inline-flex w-fit',
+          disabled && 'opacity-30',
+          className,
+        )}
         {...rest}
       >
         <div className='inline-flex'>{children}</div>
         <div
-          className={cn(
-            dotVariants({ size, shape, pin, appearance, disabled }),
-          )}
+          className={cn(dotVariants({ size, shape, pin, appearance }))}
           style={style}
         >
           <Icon size={dotIconSizeMap[size]} className='text-on-interactive' />
--- a/libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
+++ b/libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
@@ -8,6 +8,15 @@ import { useMemo } from 'react';
 import type { IconSize } from '../Icon';
 import type { DotIconPin, DotIconProps, DotIconSize } from './types';
 
+const rootVariants = cva('relative inline-flex w-fit', {
+  variants: {
+    disabled: {
+      true: 'opacity-30',
+      false: '',
+    },
+  },
+});
+
 const dotVariants = cva(
   'absolute z-10 box-content flex items-center justify-center overflow-hidden border-base-inverted',
   {
@@ -110,11 +119,7 @@ export const DotIcon = ({
     <DisabledProvider value={{ disabled: false }}>
       <div
         ref={ref}
-        className={cn(
-          'relative inline-flex w-fit',
-          disabled && 'opacity-30',
-          className,
-        )}
+        className={cn(rootVariants({ disabled, className }))}
         {...rest}
       >
         <div className='inline-flex'>{children}</div>
--- a/libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
+++ b/libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
@@ -115,7 +115,6 @@ export const DotIcon = ({
       <div
         ref={ref}
         className={cn('relative inline-flex w-fit', className)}
-        data-disabled={disabled || undefined}
         {...rest}
       >
         <div className='inline-flex'>{children}</div>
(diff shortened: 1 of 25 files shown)


libs/ui-react/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -109,35 +113,46 @@ export const DotSymbol = ({
...
-            onError={() => setError(true)}
-          />
+    <DisabledProvider value={{ disabled: false }}>
+      <div
+        ref={ref}
+        className={cn(
+          'relative inline-flex w-fit',
+          disabled && 'opacity-30',
+          className,
         )}
```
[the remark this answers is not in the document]
aquelemiguel: wrong... using outer wrapper for this otherwise with that approach we'll get double opacity such as

image

libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
```diff
@@ -92,25 +96,36 @@ export const DotIcon = ({
...
+    <DisabledProvider value={{ disabled: false }}>
       <div
-        className={cn(dotVariants({ size, shape, pin, appearance }))}
-        style={style}
+        ref={ref}
+        className={cn(
+          'relative inline-flex w-fit',
+          disabled && 'opacity-30',
+          className,
+        )}
```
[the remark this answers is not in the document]
aquelemiguel: same

libs/ui-rnative/src/lib/Components/DotSymbol/DotSymbol.tsx
```diff
@@ -118,42 +122,54 @@ export const DotSymbol = ({
...
-            />
-          )}
+    <DisabledProvider value={{ disabled: false }}>
+      <Box
+        ref={ref}
+        lx={lx}
+        style={StyleSheet.flatten([
+          { position: 'relative' },
+          disabled && { opacity: 0.3 },
+          style,
```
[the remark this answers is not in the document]
aquelemiguel: same

libs/ui-rnative/src/lib/Components/DotIcon/DotIcon.tsx
```diff
@@ -119,27 +123,39 @@ export const DotIcon = ({
...
-        <Box testID='dot-icon-dot' style={styles.dot}>
-          <Icon size={dotIconSizeMap[size]} style={styles.icon} />
+    <DisabledProvider value={{ disabled: false }}>
+      <Box
+        ref={ref}
+        lx={lx}
+        style={StyleSheet.flatten([
+          { position: 'relative' },
+          disabled && { opacity: 0.3 },
+          style,
```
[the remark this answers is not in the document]
aquelemiguel: same

libs/ui-react/src/lib/Components/MediaButton/MediaButton.tsx
```diff
@@ -90,7 +90,6 @@ export const MediaButton = ({
         triggerVariants({ size, iconType: effectiveIconType }),
         className,
       )}
-      data-disabled={disabled || undefined}
       disabled={disabled}
       {...props}
     >
```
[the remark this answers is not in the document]
aquelemiguel: this is on purpose https://github.com/LedgerHQ/lumen/pull/712/commits/8b15d1b142896a36c577d992491f7bd8418cc462

libs/ui-rnative/src/lib/Components/DotIcon/DotIcon.tsx
gamegee: [low]: could move that to the useStylesheet above, so it keep it clear separation of concerns.

+ small improvement: replace StyleSheet.flatten by array

libs/ui-rnative/src/lib/Components/DotSymbol/DotSymbol.tsx
gamegee: [low]: same

libs/ui-react/src/lib/Components/DotIcon/DotIcon.tsx
gamegee: [low]: same could apply -> move to a CVA once we start to have conditions

```tsx
cn(rootVariants({ disabled, className }))
```

It might seem overkill, but its just it make it declarative and extensible

gamegee: [mid]: I noticed that the disabled will be applied on:
- the DotSymbol pinned in a corner
- the element wrapped by the DotSymbol

I am wondering if we should not just apply it to the _DotSymbol pinned in a corner_

I can imagine a composition where the `<MediaImage>` is enabled, but the `DotSymbol` on the corner disabled.

---
TLDR; it mean we should apply disabled styles in the `dotVariants()`

gamegee: same for all other `dotX` components

aquelemiguel: > same could apply -> move to a CVA once we start to have conditions

good shout, even if a bit overkill, i think it fits a pattern we want to keep enforcing
applying!

> I am wondering if we should not just apply it to the DotSymbol pinned in a corner

i see your point, could be a cool addition!
but checked with laurine and for the time being we're disabling both, maybe in the future?

gamegee (approved):

PR #713, opened 2026-05-25
fix-reveal-path-android

gamegee:


libs/ui-rnative/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export { type SupportedLocale, Languages } from './i18n';
+export { RuntimeConstants } from './lib/utils';
```
gamegee, line 2: Exposes it from DS

aammami-ledger (approved):
Would be nice to add a ticket as we discussed for later when we have time to dig deeper.

PR #716, opened 2026-05-26
[D]DLS-736-feat(AmountDisplay): Add size variant support

zel-kass:
## Overview

- Add new size prop with t-shirt sizings 'sm', and default 'md'.
- Refactor size mapping.
- Export component as function instead of deprecated arrow function.
Screenshot 2026-05-26 at 16 16 26


gamegee (approved):
Missing version plan + update code-connect figma

PR #718, opened 2026-05-26
fix(ui-rnative): page indicator component not updating on page change

mitchellv-ledger:
**Context:**

The page indicator was not moving or updating correctly on slide change for product tour and walletV4Tour. 

**Reason for change:**
During investigation it was evident that the bug was coming from the lumen component itself rather than a LWM bug.

**Bug Details:**
useTimingConfig({ duration: 200, easing: 'easeInOut' }) returns a reference that ends up being unstable enough to be in the useEffect deps, combined with return () => cancelAnimation(...) in cleanup. Each time the effect re-runs, the cleanup cancels any in-progress withTiming before it has a chance to make visible progress. The dot's colorProgress never actually animates.

### DIFF
--- a/.nx/version-plans/version-plan-1779801734000-page-indicator.md
+++ b/.nx/version-plans/version-plan-1779801734000-page-indicator.md
@@ -1,0 +1,5 @@
+---
+'@ledgerhq/lumen-ui-rnative': patch
+---
+
+fix(ui-rnative): fix PageIndicator not reliably updating on slide change
--- a/libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.test.tsx
+++ b/libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.test.tsx
@@ -15,6 +15,12 @@ const renderWithProvider = (component: React.ReactElement) => {
   );
 };
 
+const wrapWithProvider = (component: React.ReactElement) => (
+  <ThemeProvider themes={ledgerLiveThemes} colorScheme='dark' locale='en'>
+    {component}
+  </ThemeProvider>
+);
+
 describe('PageIndicator Component', () => {
   describe('Rendering', () => {
     it('should render with required props', () => {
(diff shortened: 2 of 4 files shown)


libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.test.tsx
```diff
@@ -7,6 +8,50 @@ import { createRef } from 'react';
 import { ThemeProvider } from '../ThemeProvider/ThemeProvider';
 import { PageIndicator } from './PageIndicator';
 
+// Override the global reanimated mock for this file so that useAnimatedStyle
+// actually evaluates its callback. The regression test below needs the
+// rendered style array to reflect interpolated values, which the default
+// mock collapses to `{}`. jest.mock is hoisted to the top of the file.
+jest.mock('react-native-reanimated', () => {
```
gamegee: [mid]: we already have a setup file at the root of the `ui-rnative` library, could you add the new mocked method to it, it will prevent duplicates

mitchellv-ledger: I have added it, please check it looks correct

.nx/version-plans/version-plan-1779801734000-page-indicator.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ledgerhq/lumen-ui-rnative': patch
+---
+
+fix(PageIndicator): dot and strip styles now update on currentPage changes — replaced the animated Box wrapper with Animated.View (Box's StyleSheet.flatten was snapshotting reanimated styles) and inlined the withTiming config to stop the unstable timing-config identity from re-triggering the animation on every render
```
aquelemiguel: here we usually just tag the affected lib (`ui-rnative`/`ui-react`) instead of the component and we keep the description simple. maybe something like:

```suggestion
fix(ui-rnative): fix PageIndicator not reliably updating on slide change
```

libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.tsx
```diff
@@ -26,20 +23,13 @@ const useDotAnimation = ({
   const colorProgress = useSharedValue(isActive ? 1 : 0);
   const shrinkProgress = useSharedValue(isShrunk ? 1 : 0);
 
-  const timingConfig = useTimingConfig({
-    duration: 200,
-    easing: 'easeInOut',
-  });
```
aquelemiguel, line 32: i'm split here, are we positive this isn't stable across re-renders? i found it surprising given the hook is memoized. not too hot on this approach as it drops design tokens. if this also contributes to the issue, i'd explore a less aggressive approach

libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.tsx
```diff
@@ -26,20 +23,13 @@ const useDotAnimation = ({
...
-
   useEffect(() => {
-    colorProgress.value = withTiming(isActive ? 1 : 0, timingConfig);
-    return () => cancelAnimation(colorProgress);
```
aquelemiguel, line 36: without this cleanup, if unmounted animations keep running. granted, it's a trivial animation but does the cleanup cause any regressions?

aquelemiguel (changes requested):
thank you immensely @mitchellv-ledger for the report and subsequent fix! i can confirm the Reanimated wrap of Box was most definitely the issue, could you please check if the rest is strictly necessary?

libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.tsx
mitchellv-ledger: It was trivial as the animation is small so would be fine, I have managed to add it back in so that it only cleans up after the whole component dismounts.

libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.tsx
mitchellv-ledger: I have confirmed it was the cancelAnimation cleanup running on every effect re-run, not timingConfig instability.

mitchellv-ledger: Thank you for recommending to look into it more

libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.test.tsx
```diff
@@ -161,6 +167,81 @@ describe('PageIndicator Component', () => {
...
+    // Regression: dot/strip styles must update when currentPage changes.
+    // Two distinct bugs have produced the same broken-on-update symptom:
+    //   1. Wrapping in Animated.createAnimatedComponent(Box) — Box flattens
+    //      style arrays via StyleSheet.flatten, which snapshots reanimated
+    //      animated styles and stops updates.
+    //   2. An unstable timing-config object identity across renders,
+    //      which made the withTiming effect re-run every render and the
+    //      animation never settle.
+    // Initial render looked correct in both cases — only re-render exposed
+    // the issue, which is what the tests below exercise.
```
aquelemiguel, line 177: please either update this comment with the new findings (remove point 2?) or tbf, you can potentially remove to not clutter the tests, up to you!

mitchellv-ledger: good spot, I removed point 2 as that was the point that was still relevant and the part about the cancel animation seems unrelated to here.

libs/ui-rnative/src/lib/Components/PageIndicator/PageIndicator.test.tsx
```diff
@@ -161,6 +167,78 @@ describe('PageIndicator Component', () => {
     });
   });
 
+  describe('Re-rendering', () => {
+    // Regression: dot/strip styles must update when currentPage changes.
+    // Two distinct bugs have produced the same broken-on-update symptom:
+    //   1. Wrapping in Animated.createAnimatedComponent(Box) — Box flattens
```
aquelemiguel: just tweak this sentence "two distinct bugs" -> "one bug"
remove list order to bulletpoint maybe

aquelemiguel (approved):
just a nitpick on the comment, pre-approving
thank you again for your work @mitchellv-ledger 🥇

PR #719, opened 2026-05-26
[M]DLS-736-feat(AmountDisplay): Add size variant support

zel-kass:
## Overview

- Add new size prop with t-shirt sizings 'sm', and default 'md'.
- Refactor size mapping.
- Export component as function instead of deprecated arrow function.


gamegee (approved):
- Missing version plan
- Missing code-connect -> will do later

PR #725, opened 2026-05-28
Fix code connect to use helper text

aammami-ledger:


libs/ui-react/src/lib/Components/Icon/Icon.figma.tsx
```diff
@@ -6,6 +6,7 @@ figma.connect(
   'https://www.figma.com/design/zSkvGGiqcnhywp2l3HTHxA?node-id=5625%3A429',
   {
     imports: ["import { Placeholder } from '@ledgerhq/lumen-ui-react/symbols'"],
-    example: () => <Placeholder size={24} />,
+    // @ts-expect-error — placeholder value, replace with a valid IconSize
+    example: () => <Placeholder size='put correct size' />,
```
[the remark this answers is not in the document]
gamegee: [low]: What about `size={<define-icon-size>}` - its more dev-friendly and easily searchable by regex pattern

libs/ui-react/src/lib/Components/Icon/Icon.figma.tsx
```diff
@@ -1,11 +1,2990 @@
 import figma from '@figma/code-connect';
```
gamegee, line 1: [mid]: as discussed earlier, I would generate it from a template, I guess we can find related NodeNames (icon-names) and the Node-Id to loop over-it and generate this file.

gamegee: That being said, we can already merge that one and create another ticket.

Its just it will add some maintenance overhead, we have to make sure to minimize it, because now sync figma means:
- make sure we align on theme.js for rnative
- make sure the `Icon.figma` is up-to-date
- remove the `is-utility` (already done I think)
- ?

libs/ui-react/src/lib/Components/Spinner/Spinner.figma.tsx
```diff
@@ -6,6 +6,7 @@ figma.connect(
   'https://www.figma.com/design/JxaLVMTWirCpU0rsbZ30k7?node-id=9023-3681',
   {
     imports: ["import { Spinner } from '@ledgerhq/lumen-ui-react'"],
-    example: () => <Spinner size={40} />,
+    // @ts-expect-error — placeholder value, replace with a valid IconSize
+    example: () => <Spinner size='put correct size' />,
```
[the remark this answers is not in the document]
gamegee: [low]: I would go for `<insert-size>` as suggested, for better searchability pattern

gamegee (approved):

libs/ui-react/src/lib/Components/Icon/Icon.figma.tsx
gamegee: let me know ;)

Minimal action required would be to create a ticket

aammami-ledger: yes I will create a ticket, but again it's not a big deal if we have a code connect missing for an icon

moncefguettat:
j'adore <3

PR #726, opened 2026-05-29
feat(Amount-input):  Add new `size` and `align` prop

aammami-ledger:
image

### DIFF
--- a/libs/ui-react/src/lib/Components/AmountInput/AmountInput.tsx
+++ b/libs/ui-react/src/lib/Components/AmountInput/AmountInput.tsx
@@ -3,39 +3,114 @@ import {
   getFontSize,
   textFormatter,
   useDisabledContext,
+  useMergedRef,
 } from '@ledgerhq/lumen-utils-shared';
 import { cva } from 'class-variance-authority';
 import { useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
-import type { AmountInputProps } from './types';
+import type { AmountInputProps, AmountInputSize } from './types';
+
+/** Extra width for empty input (caret + placeholder). */
+const INPUT_WIDTH_PADDING_EMPTY = 33;
+/** Extra width when the input has a value. */
+const INPUT_WIDTH_PADDING_FILLED = 8;
+/** Extra width without currency — md size (heading-0 side bearings). */
+const INPUT_WIDTH_PADDING_NO_CURRENCY_MD = 24;
+/** Extra width without currency — sm size (heading-2 side bearings). */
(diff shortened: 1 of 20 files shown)


libs/ui-react/src/lib/Components/AmountInput/AmountInput.tsx
```diff
@@ -152,16 +159,26 @@ export const AmountInput = ({
...
+  useLayoutEffect(syncInputWidth, [syncInputWidth]);
+
+  // Re-sync whenever the mirror span resizes — this catches the font-loading
+  useEffect(() => {
+    const span = spanRef.current;
+    if (!span) return;
+    const observer = new ResizeObserver(syncInputWidth);
+    observer.observe(span);
+    return () => observer.disconnect();
+  }, [syncInputWidth]);
```
aammami-ledger, line 179: This is actually  fixes a existing bug I noticed (it's not major one because AmountInputs are default to empty in consumers app)


Image

## Problem:

AmountInput measures a hidden mirror <span> to size the input. On first render the custom font isn't loaded yet, so the browser uses a narrower fallback — the input ends up too small and the text gets clipped. 

## Fix

A ResizeObserver on the mirror span automatically re-syncs the input width whenever the span resizes — including when the real font loads and the glyphs get wider.

gamegee: On most consumers apps, AmountInput will be mounted after the font is loaded.
So its not a big concern

gamegee: Resize observer is indeed the good pattern, Did you manage to do the same on RN with a different API ?

aammami-ledger: RN do not have this issue, it's a web only issue

aquelemiguel (approved):
excellent work

libs/utils-shared/src/lib/inputFormatter/fontSizeFormatter.ts
```diff
@@ -1,16 +1,22 @@
-const MAX_FONT_SIZE = 48;
-const MIN_FONT_SIZE = 17;
-const SCALE_FACTOR = 2;
+const FONT_SIZE_BY_SIZE = {
+  md: { max: 48, min: 17, scale: 2, startAt: 0 },
+  sm: { max: 28, min: 12, scale: 1.5, startAt: 6 },
```
aammami-ledger: resize values provided by Laurine and Julien

zel-kass (approved):
some dark magic has happened here

PR #727, opened 2026-05-29
BREAKING_CHANGE: update DotIndicator sizes

aquelemiguel:
Closes [DLS-757](https://ledgerhq.atlassian.net/browse/DLS-757).

## Considerations

* As confirmed with Laurine, the new default is `md` (from `sm`) with a pixel size of `12px` (from `8px`).
* On ui-react, refactored `Avatar` to use `DotIndicator`, syncing the logic between both libs.
* The dot indicator should only show for avatars with sizes `sm` (`lg` dot) and `md` (`xl` dot). For larger avatar sizes `lg` and `xl`, even if the `showNotification` prop is present, the badge will not show. See demo below:

image

## Demo

Tested on iOS 26.5 (iPhone 17 simulator).

image

[DLS-757]: https://ledgerhq.atlassian.net/browse/DLS-757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.nx/version-plans/version-plan-1780069206903-react.md
+++ b/.nx/version-plans/version-plan-1780069206903-react.md
@@ -4,15 +4,17 @@
 
 BREAKING_CHANGE(ui-react): update DotIndicator sizes
 
-The `DotIndicator` size scale has shifted by one step. The previous scale (`xs | sm | md | lg`) has been replaced by (`sm | md | lg | xl`), and the default size is now `md` instead of `sm`. The underlying pixel sizes have also been reduced: the new `sm` is smaller than the old `sm`, so consumers that previously relied on the default will see a smaller dot unless they migrate.
+The `DotIndicator` size scale has been renamed and rescaled. The previous scale (`xs | sm | md | lg`) has been replaced by (`sm | md | lg | xl`), and the default size is now `md` instead of `sm`. The underlying token sizes have been reduced overall. The names did not simply shift down by one. Old `md` (`s14`) and old `lg` (`s16`) have no equivalent in the new scale.
 
 ### Migration
 
-Rename `size` values one step down the scale to preserve the previous visual size:
+The new sizes map to the old as follows, matched by token size:
 
-- `size="xs"` → `size="sm"`
-- `size="sm"` → `size="md"` (also the new default, it can be omitted)
-- `size="md"` → `size="lg"`
-- `size="lg"` → `size="xl"`
+- `size="xs"` (`s10`) → `size="lg"` (`s10`)
+- `size="sm"` (`s12`) → `size="xl"` (`s12`)
+- `size="md"` (`s14`) → no exact equivalent
+- `size="lg"` (`s16`) → no exact equivalent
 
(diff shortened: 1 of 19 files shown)


libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -72,6 +72,10 @@ export const Avatar = ({
 
   const resolvedAlt = alt || t('components.avatar.defaultAlt');
 
+  // dot indicator is not visible on larger sizes, regardless of the `showNotification` prop
+  const showNotification =
+    showNotificationProp && (size === 'sm' || size === 'md');
+
```
aammami-ledger, line 78: I'm not sure about this rule. I understand that we don't want consumers to add badges when the avatar size is large, but I don't think enforcing that is the Avatar component's responsibility. In my opinion, it's up to consumers to use the component correctly and follow the intended usage in Figma.

so maybe we should let the consumers add the dot themselves as we do on other components ?

.nx/version-plans/version-plan-1780069206903-react.md
```diff
@@ -0,0 +1,20 @@
+---
+'@ledgerhq/lumen-ui-react': patch
+---
+
+BREAKING_CHANGE(ui-react): update DotIndicator sizes
```
aammami-ledger: Actually it's more convenient to put the name of component then the package (so I could just copy pas this line when writing the msg I will post in slack)
```suggestion
BREAKING_CHANGE(DotIndicator): update DotIndicator sizes
```

.nx/version-plans/version-plan-1780069206903-rnative.md
```diff
@@ -0,0 +1,20 @@
+---
+'@ledgerhq/lumen-ui-rnative': patch
+---
+
+BREAKING_CHANGE(ui-rnative): update DotIndicator sizes
```
aammami-ledger:
```suggestion
BREAKING_CHANGE(DotIndicator): update DotIndicator sizes
```

zel-kass (approved):
Very clean work as always ! Pre-approving.
I just have one concern on Avatar, I was wondering about the fallback icon. Right now, in dark mode, it is barely visible because we have set a black icon :

Image

---

On Figma, a white Icon is used for both but it is only because it sits on a black background, that is something that we don't have, so I see two options to improve that.
1. We conditionaly display a white Icon depending on the theme to have something like this in dark mode :

Image

2. We align with Figma and put a dark background :

Image

Although this is a linear gradient, so we might need to sync with Laurine. This is just an open question btw, maybe I missed something, and it doesn't have anything to do with the code in this PR. 🫰🏽

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: based on discussion on the daily, created a ticket:
https://ledgerhq.atlassian.net/browse/DLS-770

aquelemiguel:
@zel-kass nice catch, thank you 🕵️ 
as it is unrelated to the current ticket, i've created https://ledgerhq.atlassian.net/browse/DLS-771

aammami-ledger (approved):
Approving this, and hopefully soon we will tackle the new tickets! 

Added chromatic label btw just to double check the changes.

PR #729, opened 2026-06-01
feat(MediaImage, DotIcon, DotSymbol): add support for 72px size 

aammami-ledger:
- Updated MediaImage to support a new size of 72px, including adjustments in stories and documentation.
- Enhanced DotIcon and DotSymbol components to include the 72px size in their respective size maps and stories.
- Updated type definitions to reflect the new size option for MediaImage.

### DIFF
--- a/libs/ui-react/src/lib/Components/MediaImage/MediaImage.tsx
+++ b/libs/ui-react/src/lib/Components/MediaImage/MediaImage.tsx
@@ -14,6 +14,7 @@ export const fontSizeMap: Record<MediaImageSize, number> = {
   48: 24,
   56: 24,
   64: 24,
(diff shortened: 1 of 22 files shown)


libs/ui-react/src/lib/Components/MediaImage/MediaImage.tsx
```diff
@@ -14,6 +14,7 @@ export const fontSizeMap: Record<MediaImageSize, number> = {
   48: 24,
   56: 24,
   64: 24,
+  72: 32,
```
aammami-ledger, line 17: to be validated by @LaurineDevergne

gamegee (approved):

PR #730, opened 2026-06-01
refactor(ui-rnative-visualization): align axis API with react-visualization + fix tick values

gamegee:
## Summary

Split out from [#728](https://github.com/LedgerHQ/lumen/pull/728) so the react-native-visualization axis work lands independently.

- `fix(charts)`: respect `xAxis.data` as the source of truth for tick values — `buildTicksData` derives ticks from the data itself instead of `scale.ticks()`, removing d3-invented intermediate ticks. The `ticks: []` short-circuit is preserved.
- `refactor(charts)`: align the axis API with `react-visualization` — consolidate axis config onto a single `BaseAxisProps` type, add a `nice` opt-out on numeric scales (default `true`), and cap `toScaledPoints` iteration at `xData.length` so a series with more points than the axis has labels no longer overflows the right edge.
- Updates the `app-sandbox-rnative` LineCharts demo accordingly.

## Test plan

- [ ] `nx test ui-rnative-visualization`
- [ ] Verify LineChart in `app-sandbox-rnative` renders ticks from `xAxis.data` and clips series that exceed the axis length

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.nx/version-plans/version-plan-1780320400001-rnative-visualization.md
+++ b/.nx/version-plans/version-plan-1780320400001-rnative-visualization.md
@@ -1,0 +1,23 @@
+---
+'@ledgerhq/lumen-ui-rnative-visualization': patch
+---
+
+fix(charts): respect `xAxis.data` as the source of truth for tick values
+
+When `xAxis.data` is provided without explicit `ticks`, `buildTicksData` now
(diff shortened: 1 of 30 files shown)


zel-kass (approved):
boss

PR #731, opened 2026-06-02
feat(ui-react): add Pagination component

aammami-ledger:
image

### DIFF
--- a/libs/ui-react/src/i18n/locales/en.json
+++ b/libs/ui-react/src/i18n/locales/en.json
@@ -50,6 +50,11 @@
       "previousPageAriaLabel": "Previous page",
       "nextPageAriaLabel": "Next page",
       "pageAriaLabel": "Page {{page}}"
+    },
+    "trend": {
+      "positiveAriaLabel": "Trend up {{value}}",
+      "negativeAriaLabel": "Trend down {{value}}",
+      "neutralAriaLabel": "Neutral trend {{value}}"
     }
   }
 }
--- a/libs/ui-react/src/i18n/locales/en.json
+++ b/libs/ui-react/src/i18n/locales/en.json
@@ -44,6 +44,12 @@
     "table": {
       "ascAriaLabel": "Sort ascending",
       "descAriaLabel": "Sort descending"
+    },
+    "pagination": {
+      "navigationAriaLabel": "Pagination",
+      "previousPageAriaLabel": "Previous page",
+      "nextPageAriaLabel": "Next page",
+      "pageAriaLabel": "Page {{page}}"
     }
   }
 }
--- a/libs/ui-react/src/lib/Components/Pagination/getPaginationRange.ts
+++ b/libs/ui-react/src/lib/Components/Pagination/getPaginationRange.ts
@@ -1,19 +1,20 @@
 export type PaginationRangeItem = number | 'ellipsis';
 
 const FIRST_PAGE = 1;
-
 const SIBLING_SIDES = 2;
+const FIXED_PAGE_SLOT_COUNT = 5;
+const BASE_CLUSTER_PAGE_COUNT = 3;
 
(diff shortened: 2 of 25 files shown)


libs/ui-react/src/i18n/locales/en.json
```diff
@@ -44,6 +44,12 @@
     "table": {
       "ascAriaLabel": "Sort ascending",
       "descAriaLabel": "Sort descending"
+    },
+    "pagination": {
+      "navigationAriaLabel": "Pagination",
+      "previousPageAriaLabel": "Previous page",
+      "nextPageAriaLabel": "Next page",
+      "pageAriaLabel": "Page {{page}}"
```
gamegee, line 52: [mid]: I don't remember, should we also translate to fr & other main languages ?

aammami-ledger: yes I will open a translation ticket.

libs/ui-react/src/i18n/locales/en.json
```diff
@@ -44,6 +44,17 @@
...
+    },
+    "trend": {
+      "positiveAriaLabel": "Trend up {{value}}",
+      "negativeAriaLabel": "Trend down {{value}}",
+      "neutralAriaLabel": "Neutral trend {{value}}"
```
aammami-ledger, line 57: not related to this PR ,but there where missing ( I have opned a [Localisation ticket](https://ledgerhq.atlassian.net/browse/L10N-30858) for all the new strings)

libs/ui-react/src/i18n/locales/en.json
aammami-ledger: https://ledgerhq.atlassian.net/browse/L10N-30858

libs/ui-react/src/lib/Components/Pagination/getPaginationRange.ts
```diff
@@ -0,0 +1,69 @@
...
+const range = (start: number, end: number): number[] =>
+  Array.from({ length: end - start + 1 }, (_, index) => start + index);
+
+/** Max page slots before collapsing */
+const getMaxVisiblePageSlots = (siblingCount: number): number =>
+  siblingCount * SIBLING_SIDES + 5;
+
+/** Page count in the start or end cluster when collapsed */
+const getCollapsedClusterPageCount = (siblingCount: number): number =>
+  3 + siblingCount * SIBLING_SIDES;
```
gamegee: [mid]: `3` and `5` are magic numbers here, define them as const so we have a clear naming of what they represents

libs/ui-react/src/lib/Components/Pagination/types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type PaginationProps = {
+  /**
+   * Current page, 1-indexed.
+   */
+  page: number;
+  /**
+   * Total number of pages.
+   */
+  totalPages: number;
```
gamegee, line 12: [mid]: explain that if `<= 0` it will return null 👍 

Note: We could also decide to return disabled arrow if the totalPage is `<= 0` making it explicit that the data passed to the pagination is invalid

gamegee (commented):
[mid]: There is a small layout drift when the current-page updates, I believe this is fixable if we define the same width for all elements. 

That being said, I don't think we can prevent layout drift if we have large numbers, but at least make it work for `0-9` and optimally for `10-99` would be a good UX improvement

https://github.com/user-attachments/assets/95334671-6cbe-4377-9cfa-5ade47cec6c8

libs/ui-react/src/lib/Components/Pagination/types.ts
aammami-ledger: I prefer the solution where we return null IMHO, so keeping it like this and adding the comment.

gamegee (approved):

PR #733, opened 2026-06-02
[D]DLS-768-feat(LineChart): domain Story example

zel-kass:
## Overview

Introduce a full implementation of the LineChart with all subcomponents in a real use case example.


libs/ui-react-visualization/src/lib/Components/LineChart/__stories__/cryptoChartData.test.ts
```diff
@@ -0,0 +1,177 @@
+import { describe, expect, it } from 'vitest';
```
aammami-ledger: hmmm not sure I understand why are we testing stories 🤔

zel-kass: Sonar was blocking the pr because of '0.00% coverage on new code'. Although I'd love to remove them, so if we can merge the PR without Sonar's approval, just tell me and I'll delete.

aammami-ledger: yes please do remove them, we can merge without Sonar's approval, we know better 👯

gamegee: We can start add exclusion `__stories__` and `__tests__` from sonar config
Same for the configuration of eslint/tsconfig

gamegee (approved):

aammami-ledger (approved):

PR #738, opened 2026-06-04
[D]DLS-724-feat(LineChart): introduce loading and empty state

zel-kass:
## Overview

Add loading and emptyLabel props on Line/CartesianChart, derive empty state from empty data + loading = true.
EmptyLabel component is an absolutely positioned HTML element so CartesianChart's container becomes relative parent.

https://github.com/user-attachments/assets/c41a5b9e-b926-4e41-aa5d-e4e9abc02b3b


libs/ui-react-visualization/src/lib/Components/CartesianChart/ChartEmptyLabel/ChartEmptyLabel.tsx
```diff
@@ -0,0 +1,28 @@
+import type { ReactNode } from 'react';
+
+type ChartEmptyLabelProps = {
+  children: ReactNode;
```
gamegee: [low]: can use `PropsWithChildren`

libs/ui-react-visualization/src/lib/Components/CartesianChart/ChartEmptyLabel/index.ts
```diff
@@ -0,0 +1 @@
+export { ChartEmptyLabel } from './ChartEmptyLabel';
```
gamegee: I would kill the barrel, as its internal only

libs/ui-react-visualization/src/lib/Components/CartesianChart/ShimmerAnimation/index.ts
```diff
@@ -0,0 +1 @@
+export { useShimmerAnimation } from './utils';
```
gamegee: Same remove the barrel (barrel = index.ts).

Multiple reasons, here is a short article about it:
If your ADHD brain doesn't want to read the full article you can still ask Claude to summarize 👍 

https://tkdodo.eu/blog/please-stop-using-barrel-files

libs/ui-react-visualization/src/lib/Components/CartesianChart/CartesianChart.tsx
```diff
@@ -60,9 +62,6 @@ export function CartesianChart({
   const resolvedWidth =
     typeof width === 'number' ? width : (measuredWidth ?? 0);
 
-  // The SVG canvas is enlarged by the overflow buffer on every side so edge
-  // content (labels, points, ticks) is not clipped, then shifted back via
-  // OVERFLOW_OFFSET so the drawing area still spans the container footprint.
```
gamegee, line 65: 🙏

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -64,6 +72,26 @@ export function LineChart({
...
+  const isInitialLoading = loading && !hasData;
+  const isEmpty = !loading && !hasData;
+  const isTransitionLoading = loading && hasData;
+  const isPlaceholder = isInitialLoading || isEmpty;
+
+  let ariaLabel: string | undefined;
+  if (isEmpty) {
+    ariaLabel = emptyLabel;
+  } else if (loading) {
+    ariaLabel = 'Loading chart';
```
gamegee: [mid]: use i18next translation key

gamegee: [low]: Plus I would define the aria-label in a functional programming fashion:

```tsx
const ariaLabel = useMemo(() => {
  if(isEmpty) {
    return emptyLabel
  }

  return loading ? t('mykey.path.loading') : ariaLabel
})
```

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -64,6 +72,26 @@ export function LineChart({
...
   ]);
 
+  const hasData = (series ?? []).some((s) => {
+    let validPoints = 0;
+    for (const value of s.data ?? []) {
+      if (value !== null) validPoints++;
+      if (validPoints >= 2) return true;
+    }
+    return false;
+  });
```
gamegee: [low]: Extract it as an utility function 👍 
Would be nice to call directly a function with right naming intent

```tsx
const hasData = hadValidSeriesData(series)
```

zel-kass: Love it, on it ! 🫰🏽

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -75,20 +103,48 @@ export function LineChart({
...
+            series?.map((s) => (
+              <Line
+                key={s.id}
+                seriesId={s.id}
+                showArea={showArea}
+                areaType={areaType}
+              />
+            ))
+          )}
+          {children}
```
gamegee: [low]: we could start to split a bit the JSX rendering, like having the `LineChartLine` and `LineChartContent` maybe as sub-component in the same file.

Sometimes it also allow to do some computation or hooks call in sub-components instead of top-level

libs/ui-react-visualization/src/lib/Components/LineChart/LineChartEmptyState/types.ts
```diff
@@ -0,0 +1,8 @@
+export type LineChartEmptyStateProps = {
+  /**
+   * Whether to animate the placeholder line with a shimmer (initial loading).
+   * When false, the placeholder is static (empty / no-data state).
+   * @default false
+   */
+  loading?: boolean;
+};
```
gamegee: [mid]: could leverage props composition like 

```tsx 
Pick<LineChartProps, 'loading'>
```

libs/ui-react-visualization/src/lib/Components/LineChart/__stories__/LineChart.stories.tsx
```diff
@@ -7,7 +7,7 @@ import {
   Trend,
```
gamegee, line 7: [mid]: The final story combine all state, but it would be nice to add specific stories about these states, so in the MDX we can emphasize how it works 👍 

I would recommend to have 
- one story `empty`
- one or two stories loading (with data and without data)

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -75,20 +103,48 @@ export function LineChart({
       axisPadding={axisPadding}
       enableScrubbing={enableScrubbing}
       onScrubberPositionChange={onScrubberPositionChange}
-      animate={animate}
+      animate={isTransitionLoading ? false : animate}
       magnetRadius={magnetRadius}
+      ariaLabel={ariaLabel}
+      ariaBusy={loading}
+      overlay={
+        isEmpty ? <ChartEmptyLabel>{emptyLabel}</ChartEmptyLabel> : undefined
```
gamegee: [low]: Empty label announced twice to screen readers (SVG aria-label + visible overlay text)

libs/ui-react-visualization/src/lib/Components/CartesianChart/hooks/useShimmerAnimation.ts
```diff
@@ -0,0 +1,35 @@
...
+   * `@keyframes` definition to inject inside an SVG `<style>` element.
+   */
+  keyframe: string;
+};
+
+/**
+ * Self-contained shimmer (opacity pulse) for SVG content. Shared by the initial
+ * loading placeholder and the transition-loading line.
+ */
+export const useShimmerAnimation = (): ShimmerAnimationResult => {
```
aammami-ledger, line 22: [mid] hmm why not call the file `useShimmerAnimation.ts` ? it's a hook , should no be in `utils.ts`

libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
```diff
@@ -45,6 +45,17 @@ export type CartesianChartProps = {
...
+   * Sets `aria-busy` on the chart SVG to signal that its content is loading.
+   * @default false
+   */
+  ariaBusy?: boolean;
+  /**
+   * HTML content rendered as an absolutely-positioned overlay above the chart
+   * SVG, within the same positioned container (so it aligns with the chart
+   * footprint). Used for e.g. an empty-state label.
+   */
+  overlay?: ReactNode;
```
aammami-ledger, line 58: [low] I would suggest a name that is more meanful like: `htmlOverlay`

libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
```diff
@@ -84,4 +84,21 @@ export type LineChartProps = {
...
+   * Whether the chart is loading. When there is no data, renders an animated
+   * shimmer placeholder line (initial loading). When data is present (e.g. a
+   * time-range change), the current line is recoloured to a muted grey and
+   * shimmers until the new `series` is provided (transition loading).
+   *
+   * To get the transition behaviour, keep passing the previous `series` while
+   * refetching; clearing it falls back to the initial placeholder.
+   * @default false
+   */
+  loading?: boolean;
```
aammami-ledger, line 96: [mid] Took me some time to understand this, I would simplfy it:

```suggestion
/**
 * Signals that new data is being fetched.
 *
 * - **No series**: renders an animated shimmer placeholder line.
 * - **Series present**: fades the current line to a muted grey and
 *   animates it until a new `series` is provided.
 *
 * @default false
 */
  loading?: boolean;
```

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -75,20 +103,48 @@ export function LineChart({
...
-        <Line
-          key={s.id}
-          seriesId={s.id}
-          showArea={showArea}
-          areaType={areaType}
-        />
-      ))}
-      {children}
+      {isPlaceholder ? (
+        <LineChartEmptyState loading={isInitialLoading} />
```
aammami-ledger: [mid] What I had in mind, is that this would be handled by the consumers, meaning: on consumer code:

```
{isLoading && !data ? (
  <LineChartEmptyState loading />
) : (
  <LineChart series={data} loading={isLoading} />
)}
```

this is more composable and less hidden logic that the consumers are not aware of. 

But with that being said, no strong push towards it, up to you.

libs/ui-react-visualization/src/lib/Components/CartesianChart/types.ts
```diff
@@ -55,7 +55,7 @@ export type CartesianChartProps = {
    * SVG, within the same positioned container (so it aligns with the chart
    * footprint). Used for e.g. an empty-state label.
    */
-  overlay?: ReactNode;
+  htmlOverlay?: ReactNode;
```
gamegee: [mid]: can't share the same name for rnative and react, is it required ?

libs/ui-react-visualization/src/lib/Components/CartesianChart/RevealClip/RevealClipDefs.tsx
```diff
@@ -19,7 +20,10 @@ export function RevealClipDefs({
   const duration = transitions?.enter?.duration ?? DEFAULT_DURATION_IN_SECONDS;
   const easing = transitions?.enter?.easing ?? DEFAULT_EASING;
 
-  const dataFingerprint = useComputeDataFingerprint({ series });
+  const dataFingerprint = useMemo(
+    () => computeDataFingerprint({ series }),
+    [series],
+  );
```
gamegee: [low]: having it as a hook was ok ?

gamegee: Ok maybe the memoization has been removed from the hook 👍 that's why you refactored it.
So either, 
1. memoize inside a hook
2. or just don't memoize at all and call the function directly

libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
```diff
@@ -105,14 +105,15 @@ export type LineChartProps = {
...
- * in sync; the sub-components receive them already resolved, hence `Required`.
+ * in sync;
  */
 type LineSeriesRenderProps = Required<
-  Pick<LineChartProps, 'series' | 'showArea' | 'areaType'>
->;
+  Pick<LineChartProps, 'showArea' | 'areaType'>
+> & {
+  series: Series[];
+};
```
gamegee: [mid]: to be confirmed, but the Pick with required should work as you initially fixed.
It should make the series working

https://www.typescriptlang.org/play/?ts=5.9.3#code/C4TwDgpgBAglC8UDeAoKUCGB+AXFAzsAE4CWAdgOZpQBGeZArgLY0REoC+KKokUAQgigAlCAEcGJIhAAmAHhgA+bgGMA9mUJQVKvIMSp0dKAEYANJyA

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -119,10 +119,13 @@ export function LineChart({
   const xAxisConfig = {
     ...defaultXAxisProps,
     ...xAxis,
+    position: xAxis?.position ?? defaultXAxisProps.position,
   };
   const yAxisConfig = {
     ...defaultYAxisProps,
     ...yAxis,
+    position: yAxis?.position ?? defaultYAxisProps.position,
+    width: yAxis?.width ?? defaultYAxisProps.width,
```
gamegee, line 125: [low]: I thought it would already handled by the destructuration of both objects? 
`...defaultYAxisProps ` then `...yAxis ` ?

gamegee (approved):

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
gamegee: Otherwise, later we could define utility fn for merging:

```tsx
const yAxisConfig = resolveDefaultProps(defaultYAxisProps(), yAxis)
```
The `resolveDefaultProps` could handle merging with advanced logic.


---
In additiona to that, I am thinking of grouping all constant and default props in one place.
I am not sure yet, but it could improve readability & maintenance.

I give you an example:
- For the default color of the Line path, this color should also be applied to the Scrubber Beacon, meaning we have to update the code in 2 places to update one same property. This break the DRY principle, and add maintenance overhead

PR #743, opened 2026-06-08
feat(Avatar): add appearance prop for background color options

aammami-ledger:
image

### DIFF
--- a/libs/ui-react/src/lib/Components/Avatar/Avatar.stories.tsx
+++ b/libs/ui-react/src/lib/Components/Avatar/Avatar.stories.tsx
@@ -85,8 +85,8 @@ export const SizeShowcase: Story = {
 
 export const AppearanceShowcase: Story = {
   render: () => (
-    <div className='inline-flex flex-col gap-8 body-2'>
-      <div className='inline-flex gap-16 rounded-lg bg-accent p-8'>
+    <div className='inline-flex flex-col gap-8'>
+      <div className='flex items-center justify-end gap-16'>
         <Avatar
           size='md'
           alt='gray fallback'
@@ -100,7 +100,7 @@ export const AppearanceShowcase: Story = {
           showNotification={false}
         />
       </div>
-      <div className='inline-flex gap-16 rounded-lg bg-accent p-8'>
+      <div className='flex items-center justify-end gap-16'>
         <Avatar
           size='md'
           src={exampleSrc}
--- a/libs/ui-react/src/lib/Components/Avatar/Avatar.stories.tsx
+++ b/libs/ui-react/src/lib/Components/Avatar/Avatar.stories.tsx
@@ -83,6 +83,43 @@ export const SizeShowcase: Story = {
   ),
 };
 
+export const AppearanceShowcase: Story = {
+  render: () => (
+    <div className='inline-flex flex-col gap-8 body-2'>
+      <div className='inline-flex gap-16 rounded-lg bg-accent p-8'>
+        <Avatar
+          size='md'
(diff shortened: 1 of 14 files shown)


libs/ui-react/src/lib/Components/Avatar/Avatar.stories.tsx
```diff
@@ -83,6 +83,43 @@ export const SizeShowcase: Story = {
   ),
 };
 
+export const AppearanceShowcase: Story = {
+  render: () => (
+    <div className='inline-flex flex-col gap-8 body-2'>
+      <div className='inline-flex gap-16 rounded-lg bg-accent p-8'>
```
gamegee: [mid]: remove bg-accent, consumers dev will figure it out 👍 (like for the button or icon-button)

aammami-ledger: ok fair enough

aammami-ledger: We should probably improve this at some point for a better understanding of transparent vs plain, as it can be confusing. 
 
image

libs/ui-rnative/src/lib/Components/Avatar/Avatar.test.tsx
```diff
@@ -33,6 +33,36 @@ describe('Avatar Component', () => {
...
+        <Avatar testID='avatar-id' />
+      </TestWrapper>,
+    );
+
+    expect(getByTestId('avatar-id').props.style.backgroundColor).toBe(
+      colors.bg.mutedTransparent,
+    );
+  });
+
+  it.each<['transparent' | 'gray', string]>([
```
gamegee, line 48: [low]: can re-use the props, 

```tsx
it.each<AvatarProps['appearance'], string>(...
```

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -39,7 +51,7 @@ const useStyles = ({ size }: { size: Size }) => {
           width: sizeMap[size].size,
           height: sizeMap[size].size,
           borderRadius: 9999,
-          backgroundColor: t.colors.bg.muted,
+          backgroundColor: backgroundColors[appearance],
```
gamegee, line 54: 👍

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -7,6 +7,7 @@ import { DotIndicator } from '../DotIndicator';
 import { Box } from '../Utility';
 import type { AvatarProps } from './types';
 
+type Appearance = NonNullable<AvatarProps['appearance']>;
```
gamegee, line 10: hmm ok

gamegee: [low]: Other solution would be:

```tsx
// Avatar.types.ts

export type AvatarAppearance = 'grey' | 'transparent'

export type AvatarProps = {
  appearance?: AvatarAppearance
}
```

So the `AvatarAppearance` is non-nullable and re-used in both places

gamegee (approved):
Few minor coms, LGTM

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
aammami-ledger: we already have same pattern on the size in the same file, that's why I followed it.

PR #744, opened 2026-06-08
fix(AmountInput): programmatically formatting the value 

aammami-ledger:


libs/ui-react/src/lib/Components/AmountInput/AmountInput.tsx
```diff
@@ -147,7 +147,20 @@ export const AmountInput = ({
   const spanRef = useRef<HTMLSpanElement>(null);
   const inputRef = useRef<HTMLInputElement>(null);
   const mergedRef = useMergedRef(ref, inputRef);
-  const [inputValue, setInputValue] = useState(value.toString());
+  const formatExternalValue = useCallback(
```
gamegee, line 150: I hadn't time to properly review the update on `AmountInput` sizes, it seems the component has grown a bit in complexity, especially the rendering part of it.

I would suggest few small improvements that would ease maintenance and readability.
Following mostly the Single Responsibility Pattern (SRP)

**1. Extract the width-measurement into a hook**
```tsx
const useAutoWidthInput = ({ inputValue, currencyText, size }: AutoWidthArgs) => {
  const spanRef = useRef<HTMLSpanElement>(null);
  const inputRef = useRef<HTMLInputElement>(null);
  const syncInputWidth = useCallback((): void => {
    if (!spanRef.current || !inputRef.current) return;
    const width = Math.ceil(
      Math.max(spanRef.current.scrollWidth, spanRef.current.offsetWidth),
    );
    const padding = getInputWidthPadding(inputValue, currencyText, size);
    inputRef.current.style.width = `${width + padding}px`;
  }, [inputValue, currencyText, size]);
  useLayoutEffect(syncInputWidth, [syncInputWidth]);
  useEffect(() => {
    const span = spanRef.current;
    if (!span) return;
    const observer = new ResizeObserver(syncInputWidth);
    observer.observe(span);
    return () => observer.disconnect();
  }, [syncInputWidth]);
  return { spanRef, inputRef };
};
```

**2. Extract the value/formatting into a hook**
```tsx
const useAmountInputValue = ({ value, onChange, formatOptions }) => {
  const format = useCallback(
    (v: string | number) => textFormatter(v.toString(), formatOptions),
    [/* formatOptions deps */],
  );
  const [inputValue, setInputValue] = useState(() => format(value));
  const [isChanging, setIsChanging] = useState(false);
  const prevValueRef = useRef(inputValue);
  useEffect(() => { setInputValue(format(value)); }, [value, format]);
  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
    const cleaned = format(e.target.value);
    setInputValue(cleaned);
    onChange({ ...e, target: { ...e.target, value: cleaned } });
    if (cleaned !== prevValueRef.current) setIsChanging(true);
    prevValueRef.current = cleaned;
  };
  return { inputValue, isChanging, setIsChanging, handleChange };
};
```

**3. Kill the left/right currency duplication**
The two <span> currency blocks are identical   except position. Extract a local component:

```tsx
const Currency = ({ size, style, children }: CurrencyProps) => (
  <span className={cn(currencyStyles({ size }), 'shrink-0')} style={style}>
    {children}
  </span>
);
```

**4. Worth evaluating: drop the mirror-span entirely**
The whole mirror-span + ResizeeObserver + padding-constant mecanism exists only to auto-size the input to its content. The CSS field-sizing: content property does this nativelyy on `input`:


```tsx
field-sizing-content // tailwind classname
```

gamegee: FYI fix (3) is something we should do a bit more on some components - @aammami-ledger @zel-kass

aammami-ledger: ok all fair points, that I can address in a separate PR though.

For the 4th point ~~not sure it's widely supported 🤔 to be checked indeed, could be nice if it works~~ Seems like it's supported by almost all browsers now! except Firefox! so maybe we can use it yes.

gamegee (approved):

PR #748, opened 2026-06-08
[D]DLS-656-fix(AmountDisplay): Fix overflow cut

zel-kass:
## Overview

Fix the overflow cut by combining overflow-x-visible and overflow-y-clip, and since these are not sufficient because the mask-fade-y still confines painting to the border box, had to add mask-no-clip. The webkit attribute is for older browsers support.


gamegee (approved):
LGTM, check Copilot just in case

PR #751, opened 2026-06-09
DLS-773/fix-switch-animation-for-lw-upgrade-0.81

gamegee:
## Description
- Fix Switch to have the animation work on LWM

## Note about the reanimated issue on LW

The root cause: on iOS with the New Architecture (Fabric), Reanimated 4 doesn't apply an animated transform to an empty leaf Animated.View. The Switch thumb was exactly that — a childless Animated.View animating only translateX — so its transform was never applied (the thumb stayed put), while the track color still changed via normal React re-render. Android wasn't affected because Fabric's view-flattening/handling differs per platform.

Every other Lumen animation worked because their animated views had real children (PageIndicator strip, AmountInput text, Spin) or were special (the absolute TabBar pill).

The fix: give the thumb's Animated.View a real child (the drawn circle) so it's no longer an empty leaf. (Neither the global Reanimated feature flag nor collapsable={false} addressed it.)

## Side Note
I am unsure yet why the animation is working on Lumen repository as we are using the New Architecture & the same libraries version.


libs/ui-rnative/src/lib/Components/Switch/Switch.tsx
```diff
@@ -1,9 +1,107 @@
...
+};
+
+const useAnimatedThumb = (checked: boolean, translate: number) => {
+  const timing = useTimingConfig({ duration: 200, easing: 'easeInOut' });
+
+  // Reanimated best practice: derive the animation reactively by calling
+  // `withTiming` inside `useAnimatedStyle`. When `checked` (or `translate`)
+  // changes the worklet re-runs and animates from the current value to the
+  // new target -- no shared value or effect required.
+  return useAnimatedStyle(
```
gamegee, line 98: I keep this comment as otherwise it might silently fails on LW

iqbalibrahim-ledger (approved):

zel-kass (approved):
LGTM, nice one !

PR #754, opened 2026-06-09
fix/minor-fix-bottom-sheet

gamegee:


aammami-ledger (approved):

aammami-ledger (commented):
Pre-approving this, but just to double check UI wise we did not introduce an regressions, Can we please add a screenshot of the BottomSheet in the PR description when no title no back and no close button ?

PR #755, opened 2026-06-10
feat(DataTable,Table): add column width management with DataTableColGroup

aammami-ledger:
Solution:  Introduced `DataTableColGroup` to define column widths via `<colgroup>`, ensuring proper width handling even with hidden headers or colSpan groups.

## Before:
Screenshot 2026-06-10 at 09 02 10

## After:
Screenshot 2026-06-10 at 09 02 04


libs/ui-react/src/lib/Components/DataTable/DataTable.tsx
```diff
@@ -91,6 +91,43 @@ export const DataTableRoot = <TData extends RowData>({
...
+                  md: 'hidden md:table-column',
+                  lg: 'hidden lg:table-column',
+                  xl: 'hidden xl:table-column',
+                }[meta.hideBelow],
+            )}
+          />
+        );
+      })}
+    </colgroup>
+  );
```
gamegee, line 121: [mid]: as other table component, you should first define the UI abstraction on the simple table. And then compose DataTable that will consume tanstack + the UI part.

So the `hideBelow` should be handled on the `TableCol` ui component.

I guess also you want to expose both `TableCol` & `TableColGroup` as ui bricks

aammami-ledger: Done

gamegee (commented):
missing version plan

libs/ui-react/src/lib/Components/Table/Table.tsx
```diff
@@ -107,6 +109,60 @@ export const Table = ({ children, className, ref, ...props }: TableProps) => {
...
+ */
+export const TableColGroup = ({
+  children,
+  className,
+  ref,
+  ...props
+}: TableColGroupProps) => {
+  return (
+    <colgroup ref={ref} className={className} {...props}>
+      {children}
```
gamegee: [mid]: From Claude

hideBelow on <col> won't actually collapse the column (7/10) — libs/ui-react/src/lib/Components/Table/Table.tsx:112-121

The colVariants use display-based utilities (hidden / xs:table-column), but per the CSS Tables spec, display does not apply to table-column boxes — only width, border, background, and visibility do. With table-fixed + an explicit <col> width, the column keeps reserving its width below the breakpoint, leaving an empty gap — the same class of bug this PR is trying to fix. Already flagged by Copilot. 

The spec-correct approach is visibility: collapse (e.g. collapse xs:visible), though browser support is historically inconsistent and should be validated.

libs/ui-react/src/lib/Components/DataTable/DataTable.tsx
```diff
@@ -91,6 +93,34 @@ export const DataTableRoot = <TData extends RowData>({
   );
 };
 
+/**
```
gamegee, line 96: [mid]: On the `DataTable.types.ts` also explain that the `meta.className`

libs/ui-react/src/lib/Components/DataTable/DataTable.test.tsx
```diff
@@ -97,6 +106,83 @@ describe('DataTable', () => {
...
+    expect(cols[1]?.className).toBe('hidden lg:table-column');
+  });
+
+  it('should apply both meta.className and meta.hideBelow to col elements', () => {
+    const columns: ColumnDef<TestData>[] = [
+      { accessorKey: 'name', header: 'Asset' },
+      {
+        accessorKey: 'price',
+        header: 'Price',
+        meta: { hideBelow: 'md' as const, className: 'w-52' },
```
gamegee: [low]: `w-52` doesn't exist

libs/ui-react/src/lib/Components/Table/Table.stories.tsx
```diff
@@ -205,6 +205,10 @@ export const AppearanceShowcase: Story = {
 };
 
 export const ResponsiveLayout: Story = {
+  parameters: {
+    chromatic: { viewports: [800, 1280] },
```
aammami-ledger, line 211: this is cool fyi @gamegee so we can test responsiveness

libs/ui-react/src/lib/Components/Table/Table.tsx
aammami-ledger: ok so after some back and forth I ended up removing hideBelow from the newly added `TableCol`
I don’t fully remember why I originally added it, I believe it was to FIX a responsiveness regression 🤔 . 

However, I’m currently unable to reproduce that issue. Everything now appears to be working as expected, with hideBelow on both `TableCell `and `TableHeaderCell`.

libs/ui-react/src/lib/Components/Table/Table.stories.tsx
gamegee: noice

libs/ui-react/src/lib/Components/Table/Table.tsx
gamegee: Ok maybe as suggested, we don't have the same behavior on several browsers ?

aammami-ledger: yes I ve tested and indeed, not same behaviour. But i've removed it now so all good.

gamegee (approved):

PR #759, opened 2026-06-12
DLS-783/fix-offset-issues-charts-rnative

gamegee:


apps/app-sandbox-rnative/src/app/blocks/LineCharts.tsx
```diff
@@ -489,6 +489,12 @@ const RandomAutoUpdate = () => {
     return () => clearInterval(id);
   }, []);
 
+  const data = series[0].data;
+  const min = Math.min(...data);
+  const max = Math.max(...data);
+  const minIdx = data.indexOf(min);
+  const maxIdx = data.indexOf(max);
```
gamegee, line 496: just adding some point in the sandbox to check animations

libs/ui-rnative-visualization/src/lib/Components/Point/Point.tsx
```diff
@@ -50,7 +53,7 @@ export function Point({
 }: Readonly<PointProps>) {
   const { getXScale, getYScale, getXAxisConfig, drawingArea } =
     useCartesianChartContext();
-  const clipPath = useRevealClip();
+  const fadeProps = useRevealFadeProps();
```
gamegee, line 56: 2nd animation hook, that is now used for the `<Point>`, but can be used for other elements if needed later

libs/ui-rnative-visualization/src/lib/Components/LineChart/LineChartEmptyState/LineChartEmptyState.tsx
```diff
@@ -27,7 +27,7 @@ export function LineChartEmptyState({
   const { drawingArea } = useCartesianChartContext();
   const { theme } = useTheme();
   const { animatedProps } = useShimmerAnimation(loading);
-  const clipPath = useRevealClip();
+  const clipPath = usePathReveal();
```
gamegee, line 30: 1st animation is just renamed but remains the same

libs/ui-rnative-visualization/src/lib/Components/CartesianChart/RevealAnimation/RevealAnimationProvider.test.tsx
```diff
@@ -71,26 +76,45 @@ describe('RevealClipDefs', () => {
...
+      <RevealAnimationProvider
+        drawingArea={drawingArea}
+        series={series}
+        transitions={{ enter: { duration: 0.5 } }}
+      >
+        <PointConsumer />
+      </RevealAnimationProvider>,
+    );
+
+    const consumer = UNSAFE_root.findByProps({ testID: 'point-consumer' });
```
zel-kass, line 105: [low]: why do we have to use an UNSAFE method here ? Isn't there any other way to achieve the same result ?

zel-kass (approved):
Beautiful 🦾

PR #760, opened 2026-06-12
[D]DLS-784-feat(LineChart): Add null/missing data support

zel-kass:
## Overview

Introduce null data handling with the `connectNulls` prop. It can either be passed to `LineChart` (chart-wide) or specified on an individual `series` object (per-series).

By default, `null` entries in a series' `data` create **gaps** in the line and area. With `connectNulls`, those nulls are skipped so the line is drawn **continuously across the gap**.

Screenshot 2026-06-12 at 16 56 16

## Resolution order
`connectNulls` is resolved with the following precedence (first defined wins):
1. `connectNulls` on `LineChart` — chart-wide override applied to every line.
2. `connectNulls` on the `series` object — per-series setting.
3. `false` (default) — nulls render as gaps.
## Changes
- **`connectNulls` on `LineChart`** (`LineChartProps`): chart-wide override. When set, it applies to every line and overrides any per-series value.
- **`connectNulls` on `series`** (`Series`): per-series opt-in, used when no chart-level value is provided.
- **`Line`**: resolves `connectNulls` (prop → series → `false`) and forwards it to the point projection. Threaded through `LineChartContent`, `LineChartLines`, and `LineChartTransitionLines`.
- **`toScaledPoints`**: when `connectNulls` is `true`, non-finite values (e.g. `null`) are skipped entirely; when `false`, they are kept as `[x, null]` holes so the line/area generators break into gaps via their `.defined()` check.
- **Axis `showLabels`** (`BaseAxisProps`): new prop (default `true`) to hide tick text labels while keeping ticks/grid, on both `XAxis` and `YAxis`.

### DIFF
--- a/libs/ui-react-visualization/src/lib/Components/Axis/XAxis/XAxis.stories.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/Axis/XAxis/XAxis.stories.tsx
@@ -90,6 +90,19 @@ export const TickLabelFormatter: Story = {
   },
 };
 
+/**
+ * `showLabels={false}` hides the tick text while keeping grid lines and tick
+ * marks, useful when the values are conveyed elsewhere (e.g. a scrubber tooltip).
+ */
+export const HideLabels: Story = {
+  args: {
+    data: monthLabels,
+    showGrid: true,
+    showTickMark: true,
+    showLabels: false,
+  },
+};
+
 /** `ticks` pins explicit positions instead of the auto-computed set. */
 export const CustomTicks: Story = {
   args: {
--- a/libs/ui-react-visualization/src/lib/Components/Line/types.ts
+++ b/libs/ui-react-visualization/src/lib/Components/Line/types.ts
@@ -30,4 +30,11 @@ export type LineProps = {
    * @default 'bump'
    */
   curve?: CurveType;
+  /**
+   * When true, skips null values and draws a continuous line across gaps.
+   * When false, null values create gaps in the line.
+   * When omitted, falls back to the `connectNulls` defined on the series.
+   * @default false
+   */
+  connectNulls?: boolean;
 };
--- a/libs/ui-react-visualization/src/lib/Components/LineChart/__stories__/LineChart.stories.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/LineChart/__stories__/LineChart.stories.tsx
@@ -19,6 +19,9 @@ import {
   CHART_HEIGHT,
   CHART_WIDTH,
   curveSeries,
+  formatScrubberValue,
+  missingDataPages,
+  missingDataSeries,
   monthLabels,
   multiSeries,
   sampleSeries,
@@ -36,13 +39,6 @@ import {
   type Period,
   usdFormatter,
 } from './cryptoChartData';
-
-const integerFormatter = new Intl.NumberFormat('en-US', {
-  maximumFractionDigits: 0,
-});
-
-const formatScrubberValue = (value: number | null): string =>
-  value === null ? '—' : integerFormatter.format(value);
 
 const meta = {
   component: LineChart,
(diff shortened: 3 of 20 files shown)


libs/ui-react-visualization/src/lib/Components/Axis/XAxis/XAxis.stories.tsx
```diff
@@ -90,6 +90,19 @@ export const TickLabelFormatter: Story = {
   },
 };
 
+/**
+ * `showLabels={false}` hides the tick text while keeping grid lines and tick
+ * marks, useful when the values are conveyed elsewhere (e.g. a scrubber tooltip).
+ */
+export const HideLabels: Story = {
```
gamegee, line 97: 🫶 le boss

libs/ui-react-visualization/src/lib/Components/Line/types.ts
```diff
@@ -30,4 +30,11 @@ export type LineProps = {
    * @default 'bump'
    */
   curve?: CurveType;
+  /**
+   * When true, skips null values and draws a continuous line across gaps.
+   * When false, null values create gaps in the line.
+   * When omitted, falls back to the `connectNulls` defined on the series.
+   * @default false
+   */
+  connectNulls?: boolean;
```
gamegee, line 39: [low]: Ok, I was wondering if we wanted to expose it to consumers, but why not - I try to see if applying connectNull is really a usecase.

LGTM anyway

libs/ui-react-visualization/src/lib/Components/LineChart/__stories__/LineChart.stories.tsx
```diff
@@ -37,6 +37,13 @@ import {
   usdFormatter,
 } from './cryptoChartData';
 
+const integerFormatter = new Intl.NumberFormat('en-US', {
+  maximumFractionDigits: 0,
+});
+
+const formatScrubberValue = (value: number | null): string =>
+  value === null ? '—' : integerFormatter.format(value);
```
gamegee: [low]: at some point we could move all utils in the other stories file

libs/ui-react-visualization/src/lib/Components/LineChart/__stories__/LineChart.stories.tsx
```diff
@@ -105,19 +112,73 @@ export const CustomLine: Story = {
...
+  render: () => {
+    const pages = [
+      'Page A',
+      'Page B',
+      'Page C',
+      'Page D',
+      'Page E',
+      'Page F',
+      'Page G',
+    ];
```
gamegee: 👍 encore trop fort

libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
```diff
@@ -21,6 +21,12 @@ export type LineChartProps = {
    * @default 'gradient'
    */
   areaType?: 'gradient';
+  /**
+   * Chart-wide override for null handling. When set, it applies to every line
+   * and overrides any per-series `connectNulls`. When omitted, each series
+   * follows its own `connectNulls` (default `false`, i.e. gaps at nulls).
+   */
```
gamegee, line 28: [mid]: add `@default false`

gamegee: Sidenote, should it be at the Series level, like we do now for `stroke` ?

libs/ui-react-visualization/src/lib/utils/types.ts
```diff
@@ -61,6 +61,13 @@ export type Series = {
    * @default 'bump'
    */
   curve?: CurveType;
+  /**
+   * When true, skips null values and draws a continuous line across gaps.
+   * When false (default), null values create gaps in the line.
+   * Overridden by the chart-level `connectNulls` prop when that is set.
+   * @default false
+   */
+  connectNulls?: boolean;
```
gamegee, line 70: [mid]: the property definition + comment description of `connectNulls` is duplicated multiple times across the codebase.

Its not the only candidate, but probably at some point we should reduce the duplication of types as well

gamegee: It can be a follow-up ticket where we identify quality improvements on duplicates, code discovery & structure 👍

libs/ui-react-visualization/src/lib/Components/LineChart/types.ts
zel-kass: Can't put default to false, it would break the resolving chain and would omit the per series connectNulls subscription intended behaviour.

About the Sidenote, this is already the case, this behaviour is implemented, the per series subscription overrides any other parents connectNulls.

gamegee (approved):

PR #765, opened 2026-06-16
DLS-735/benchmark-perf-and-improvements

gamegee:
Benchmark ran locally with throttle 4x slowdown

Metric | 1k | 10k | 100k
-- | -- | -- | --
Cold render (scripting ms) | 2100ms | 2400ms | 4500ms
DOM nodes | 9400 | 90400 | 900400
Layout+paint ms | System 3,088 ms Scripting 2,754 ms Rendering 56 ms Painting 5 ms Loading 5 ms | Scripting 3,989 ms System 3,389 ms Rendering 26 ms Loading 4 ms Painting 4 ms |  
Scrub FPS | 67 fps | 27 fps |  
Re-renders per scrub | All point, but memoized | All point but memoized |  
Heap growth / detached nodes | No leak | No leak |


libs/ui-react-visualization/src/lib/Components/Point/pointContext/MagneticPointsProvider.tsx
```diff
@@ -32,14 +34,24 @@ export function MagneticPointsProvider({
...
 
   return (
-    <MagneticPointsContextProvider value={value}>
-      {children}
-    </MagneticPointsContextProvider>
+    <MagneticRegistryProvider value={registry}>
+      <MagneticSnapshotProvider value={snapshot}>
+        {children}
+      </MagneticSnapshotProvider>
+    </MagneticRegistryProvider>
```
gamegee, line 55: ## Improvement 1

**TLDR;**  

The "two providers" split the magnetic feature into a write API (stable, for the many) and a read snapshot (volatile, for the one), which is what keeps both mount and scrub render budgets flat.

---
**Details**

The actors: 

- Many Points — each magnetic Point registers itself as a snap target. They are writers to the registry, and there can be dozens of them.
- One ScrubberProvider — it reads the set of snap targets to compute magnetization while scrubbing. It is the single reader.

The whole design goal is: writers (Points) must never re-render when the set of magnetic points changes; only the single reader (Scrubber) should react to that change.

libs/ui-rnative-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -127,17 +127,24 @@ export const LineChart = ({
...
+
+  const yAxisConfig = useMemo(
+    () => ({
+      ...defaultYAxisProps,
+      ...yAxis,
+      position: yAxis?.position ?? defaultYAxisProps.position,
+      width: yAxis?.width ?? defaultYAxisProps.width,
+    }),
+    [yAxis],
+  );
```
gamegee, line 147: ## Improvement 2

Make sure the axis definition is stable, preventing unecessary costly recomputation of axis.

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.performance.test.tsx
```diff
@@ -0,0 +1,136 @@
+import { act, fireEvent, render } from '@testing-library/react';
```
gamegee, line 1: ## Improvement 3

Add performance budget, that will prevent regression on that side of thing.

The tests focus on specific area where perf can be an issue: Points & Scrubber updates

zel-kass (approved):
lgtm

PR #768, opened 2026-06-16
[D]DLS-809-feat(Point): clamp Point label inside drawing area

zel-kass:
## Overview
Add clamping logic to the `Point` label so it always stays inside the drawing
area, even when the point sits at the chart's edges. Previously, a label
centered on an edge point would overflow the chart horizontally.
When a centered label would overflow, it is anchored to the edge and grows
inward; the arrow keeps pointing at the exact data point. The behavior is
controlled by a new `labelAlignment` prop (default `true`) so consumers can
opt out.
Also refactored `Point` to separate concerns:
- Extracted chart/behavior logic into a `usePointGeometry` hook (projection,
  magnetic registration, reveal style, in-bounds gating).
- Split the label computation into content (`resolveLabel`) and geometry
  (`computeLabelGeometry`).
- Kept pure helpers in `utils.ts` and React hooks alongside the component.

---

Screenshot 2026-06-16 at 20 29 26

### DIFF
--- a/libs/ui-react-visualization/src/lib/Components/Point/Point.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/Point/Point.tsx
@@ -73,7 +73,7 @@ export function Point({
   size = DEFAULT_SIZE,
   onClick,
   magnetic = false,
-  clampLabelToBounds = true,
+  labelAlignment = 'auto',
 }: Readonly<PointProps>) {
   const { pixel, drawingArea, revealStyle, isVisible } = usePointGeometry({
     dataX,
@@ -96,7 +96,7 @@ export function Point({
         labelPosition,
         showLabelArrow,
         area: drawingArea,
-        clamp: clampLabelToBounds,
+        alignment: labelAlignment,
       })
     : null;
 
--- a/libs/ui-react-visualization/src/lib/Components/Point/Point.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/Point/Point.tsx
@@ -1,6 +1,6 @@
 import { cssVar } from '@ledgerhq/lumen-design-core';
 
-import { DEFAULT_SIZE, STROKE_WIDTH } from './constants';
+import { DEFAULT_SIZE, LABEL_FONT_SIZE, STROKE_WIDTH } from './constants';
 import type {
   PointArrowProps,
   PointLabelProps,
@@ -22,7 +22,7 @@ export function PointLabel({
       dominantBaseline={dominantBaseline}
       style={{
         fill: cssVar('var(--text-base)'),
-        fontSize: cssVar('var(--font-style-body-4-size)'),
+        fontSize: LABEL_FONT_SIZE,
         fontWeight: cssVar('var(--font-style-body-4-weight-medium)'),
         fontFamily: cssVar('var(--font-family-font)'),
         ...style,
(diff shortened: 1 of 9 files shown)


libs/ui-react-visualization/src/lib/Components/Point/utils.ts
```diff
@@ -63,22 +68,33 @@ export const resolveDataXToIndex = (
...
+  const halfWidth =
+    (label.length * LABEL_FONT_SIZE * LABEL_CHAR_WIDTH_RATIO) / 2;
+  const arrowOffset = hasArrow ? ARROW_WIDTH / 2 : 0;
+
+  if (pixelX - halfWidth < area.x) {
+    return { x: pixelX - arrowOffset, textAnchor: 'start' };
+  }
+  if (pixelX + halfWidth > area.x + area.width) {
+    return { x: pixelX + arrowOffset, textAnchor: 'end' };
+  }
```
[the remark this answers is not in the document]
zel-kass: This is a known and intended behaviour.

libs/ui-react-visualization/src/lib/Components/Point/utils.ts
```diff
@@ -63,22 +69,33 @@ export const resolveDataXToIndex = (
...
+  alignment: LabelAlignment,
+  hasArrow: boolean,
+): { x: number; textAnchor: LabelTextAnchor } => {
+  if (alignment === 'center') {
+    return { x: pixelX, textAnchor: 'middle' };
+  }
+
+  const halfWidth =
+    (label.length * LABEL_FONT_SIZE * LABEL_CHAR_WIDTH_RATIO) / 2;
+  const arrowOffset = hasArrow ? ARROW_WIDTH / 2 : 0;
```
gamegee, line 90: [mid]:

LABEL_FONT_SIZE is the hardcoded 10, but the default PointLabel actually renders with the --font-style-body-4-size token. If that token isn't 10px, the overflow decision is computed against the wrong size.

-> Maybe use the constant as well in the Point JSX instead of a token

libs/ui-react-visualization/src/lib/Components/Point/constants.ts
```diff
@@ -4,3 +4,4 @@ export const ARROW_WIDTH = 6;
 export const ARROW_HEIGHT = 4;
 export const GAP = 4;
 export const LABEL_FONT_SIZE = 10;
+export const LABEL_CHAR_WIDTH_RATIO = 0.6;
```
gamegee, line 8: [low]: see to be a magic number, we are expecting a font with monospace.
Probably worth a small comment that explain its an average ratio

libs/ui-react-visualization/src/lib/Components/Point/Point.tsx
```diff
@@ -91,28 +112,23 @@ export function Point({
...
-          strokeWidth={STROKE_WIDTH}
-        />
+        <PointMarker x={pixel.x} y={pixel.y} size={size} color={color} />
       )}
-      {renderArrow && (
-        <polygon
-          data-testid='point-arrow'
-          points={buildArrowPoints(pixel.x, pixel.y, radius, labelPosition)}
-          style={{ fill: cssVar('var(--text-base)') }}
+      {labelGeometry?.renderArrow && (
```
gamegee: [low]: this is just mirroring `showLabelArrow`, consider using it directly and don't return `renderArrow` from `computeGeometry` fn

gamegee (commented):
few coms,
Not sure if we need a story for that, probably not

gamegee (approved):

PR #773, opened 2026-06-19
BREAKING_CHANGE: add type-safety for compound components

aquelemiguel:
Closes [DLS-811](https://ledgerhq.atlassian.net/browse/DLS-811).

Based on the suggestions found in this article:
https://tkdodo.eu/blog/building-type-safe-compound-components#component-factory-pattern

## Considerations

Tagged as BREAKING_CHANGE because (from the version plan):

> The generics on `Select` components now take the value union `T` as the first parameter, followed by `TMeta` (previously `TMeta` was the only parameter).

Same for `OptionList`. This is a type-level breaking change only, so IMO this has very small surface. Still, just to bring extra attention to it, I've tagged it as a breaking change.

[DLS-811]: https://ledgerhq.atlassian.net/browse/DLS-811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-react/src/lib/Components/SegmentedControl/SegmentedControl.tsx
+++ b/libs/ui-react/src/lib/Components/SegmentedControl/SegmentedControl.tsx
@@ -1,5 +1,6 @@
 import { cn, useDisabledContext } from '@ledgerhq/lumen-utils-shared';
 import { cva } from 'class-variance-authority';
+import type { ReactElement } from 'react';
 import { useRef } from 'react';
 import {
   SegmentedControlContextProvider,
@@ -8,6 +9,7 @@ import {
 import type {
   SegmentedControlButtonProps,
   SegmentedControlProps,
+  SegmentedControlValue,
 } from './types';
 import {
   usePillElementLayoutEffect,
@@ -64,7 +66,9 @@ const segmentedControlStyles = {
   ),
 };
 
-export function SegmentedControlButton({
+export function SegmentedControlButton<
+  T extends SegmentedControlValue = SegmentedControlValue,
+>({
   value,
   children,
   icon: Icon,
@@ -72,7 +76,7 @@ export function SegmentedControlButton({
   onClick,
   className,
   ...props
-}: SegmentedControlButtonProps) {
+}: SegmentedControlButtonProps<T>) {
   const { selectedValue, onSelectedChange, disabled, tabLayout } =
     useSegmentedControlContext();
   const selected = selectedValue === value;
@@ -106,7 +110,9 @@ export function SegmentedControlButton({
   );
 }
 
-export function SegmentedControl({
+export function SegmentedControl<
+  T extends SegmentedControlValue = SegmentedControlValue,
+>({
   selectedValue,
   onSelectedChange,
   children,
@@ -115,7 +121,7 @@ export function SegmentedControl({
   appearance = 'background',
   tabLayout = 'fixed',
   ...props
-}: SegmentedControlProps) {
+}: SegmentedControlProps<T>) {
   const disabled = useDisabledContext({
     consumerName: 'SegmentedControl',
     mergeWith: { disabled: disabledProp },
@@ -134,7 +140,12 @@ export function SegmentedControl({
 
   return (
     <SegmentedControlContextProvider
-      value={{ selectedValue, onSelectedChange, disabled, tabLayout }}
+      value={{
+        selectedValue,
+        onSelectedChange: onSelectedChange as (value: string) => void,
+        disabled,
+        tabLayout,
+      }}
     >
       <div
         {...props}
@@ -162,4 +173,15 @@ export function SegmentedControl({
       </div>
     </SegmentedControlContextProvider>
   );
+}
+
+export function createSegmentedControl<
+  T extends SegmentedControlValue = never,
+>(): {
+  SegmentedControl: (props: SegmentedControlProps<T>) => ReactElement;
+  SegmentedControlButton: (
+    props: SegmentedControlButtonProps<T>,
+  ) => ReactElement;
+} {
+  return { SegmentedControl, SegmentedControlButton };
 }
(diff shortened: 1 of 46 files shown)


libs/ui-react/src/lib/Components/SideBar/SideBar.tsx
```diff
@@ -98,18 +100,18 @@ export const SideBar = ({
...
     onChange: onCollapsedChange,
   });
 
-  const [active, setActive] = useControllableState({
+  const [active, setActive] = useControllableState<string>({
     prop: controlledActive,
     defaultProp: defaultActive ?? '',
-    onChange: onActiveChange,
+    onChange: onActiveChange as ((value: string) => void) | undefined,
   });
```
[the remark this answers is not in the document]
aquelemiguel: fair

libs/ui-react/src/lib/Components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -134,7 +140,12 @@ export function SegmentedControl({
 
   return (
     <SegmentedControlContextProvider
-      value={{ selectedValue, onSelectedChange, disabled, tabLayout }}
+      value={{
+        selectedValue,
+        onSelectedChange: onSelectedChange as (value: string) => void,
+        disabled,
+        tabLayout,
+      }}
```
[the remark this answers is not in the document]
aquelemiguel: fair

libs/ui-rnative/src/lib/Components/TabBar/TabBar.tsx
```diff
@@ -256,12 +256,13 @@ export function TabBar({
...
-  }
-
   return (
-    <TabBarContextProvider value={{ active, onTabPress: handleTabPress }}>
+    <TabBarContextProvider
+      value={{
+        active,
+        onTabPress: onTabPress as (value: string) => void,
+      }}
+    >
```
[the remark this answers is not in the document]
aquelemiguel: fair

libs/ui-react/src/lib/Components/Select/types.ts
```diff
@@ -8,9 +8,14 @@ import type { SearchInputProps } from '../SearchInput/types';
 
 export type MetaShape = Record<string, unknown>;
 
-export type SelectItemData<TMeta extends MetaShape = MetaShape> = {
+export type SelectValue = string;
+
+export type SelectItemData<
+  T extends SelectValue = SelectValue,
+  TMeta extends MetaShape = MetaShape,
+> = {
```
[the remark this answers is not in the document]
aquelemiguel: kept as patch but tagged as breaking change because of this one issue

gamegee (commented):
Its noted as a BREAKING_CHANGE, but do we agree that the elements are still usable without the typesafe factory ?

libs/ui-rnative/src/lib/Components/SegmentedControl/SegmentedControl.mdx
```diff
@@ -87,6 +89,33 @@ export default function Example() {
...
+
+function Example() {
+  const [view, setView] = React.useState<View>('preview');
+
+  return (
+    <SegmentedControl selectedValue={view} onSelectedChange={setView}>
+      <SegmentedControlButton value='preview'>Preview</SegmentedControlButton>
+      <SegmentedControlButton value='raw'>Raw</SegmentedControlButton>
+      <SegmentedControlButton value='blame'>Blame</SegmentedControlButton>
+      {/* <SegmentedControlButton value='wrong' /> ts(2322) */}
```
gamegee: ✂️

libs/ui-rnative/src/lib/Components/SegmentedControl/SegmentedControl.stories.tsx
```diff
@@ -170,3 +174,32 @@ export const WithTrailingContent: Story = {
     );
   },
 };
+
+type View = 'preview' | 'raw' | 'blame';
+const TypedSegmentedControl = createSegmentedControl<View>();
```
gamegee, line 179: [mid]: what about destructuring directly from line 179 ?, so the JSX output is way simpler.
Also, should we just go typesafe as the default way of doing stories ?

libs/ui-rnative/src/lib/Components/SegmentedControl/SegmentedControl.stories.tsx
```diff
@@ -170,3 +174,32 @@ export const WithTrailingContent: Story = {
...
+
+export const Typed: Story = {
```
gamegee: [low]: If you decide to keep a "typed" story, I would suggest to go for `Typesafe factory" as the name of the story

aquelemiguel:
> Its noted as a BREAKING_CHANGE, but do we agree that the elements are still usable without the typesafe factory ?

@gamegee 100%, you can still use the old pattern.
Only a breaking change for consumers who were passing the meta type.

libs/ui-rnative/src/lib/Components/SegmentedControl/SegmentedControl.stories.tsx
aquelemiguel: i like it, more understandable

libs/ui-rnative/src/lib/Components/SegmentedControl/SegmentedControl.stories.tsx
aquelemiguel: > what about destructuring directly from line 179 ?, so the JSX output is way simpler.

destructuring conflicts with the existing import, unless i alias it.
inclined to think this is a bit cleaner IMO

> Also, should we just go typesafe as the default way of doing stories ?

i'm thinking since it's an unique pattern most consumers won't be accustomed to, we still default to the standard. likewise it's important for us to verify the old pattern still works.

but internally when using these compounds as building blocks for other components or in the sandbox we should def use the factory, yeah.

gamegee: Lets have a story for typesafety then it makes sense, I am thinking maybe of putting the implementation snipped directly inside the Overview panel - As my goal is just to ditch the implementation panel at some point

gamegee (approved):

aammami-ledger (approved):

PR #775, opened 2026-06-22
chore(Chromatic): add pauseAnimationAtEnd parameter to pause animation

aammami-ledger:
## Fix Chromatic false positives on animated components
Chromatic was flagging spinner and loading-button stories as visual changes on every run due to non-deterministic animation frame capture.

React (ui-react): — CSS animations (animate-spin): 
- Added chromatic: `{ pauseAnimationAtEnd: true}` to Spinner stories and to the Loading/StatesShowcase Button stories.

React Native (ui-rnative) — Reanimated JS animations can't be paused by Chromatic directly, so two changes:

- Added `ReduceMotion.System` to all infinite withRepeat calls in Spin and Pulse. Without it, Reanimated ignores any global reduced-motion config — **and also ignores the device's accessibility "Reduce Motion" setting.
- Added a global Storybook decorator in preview.tsx that renders `<ReducedMotionConfig mode={ReduceMotion.Always} />` when `isChromatic()` is `true`, freezing all opted-in animations at their first frame.

------

Link to doc: https://www.chromatic.com/docs/animations/

------
So now on snapshots no animations (always initial state): 

image

### DIFF
--- a/libs/ui-react/src/lib/Components/Button/Button.stories.tsx
+++ b/libs/ui-react/src/lib/Components/Button/Button.stories.tsx
@@ -86,6 +86,7 @@ export const Loading: Story = {
     loading: true,
   },
   parameters: {
+    chromatic: { pauseAnimationAtEnd: true },
     docs: {
       source: {
         code: `
@@ -154,6 +155,9 @@ export const SizesShowcase: Story = {
 };
 
 export const StatesShowcase: Story = {
+  parameters: {
+    chromatic: { pauseAnimationAtEnd: true },
+  },
   render: () => (
     <div className='flex items-center gap-4'>
       <Button appearance='base'>Default</Button>
--- a/libs/ui-rnative/.storybook/preview.tsx
+++ b/libs/ui-rnative/.storybook/preview.tsx
@@ -1,5 +1,6 @@
 import type { Preview } from '@storybook/react-native-web-vite';
-
+import isChromatic from 'chromatic/isChromatic';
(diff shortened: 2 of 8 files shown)


libs/ui-react/src/lib/Components/Button/Button.stories.tsx
```diff
@@ -85,6 +85,7 @@ export const Loading: Story = {
     loading: true,
   },
   parameters: {
+    chromatic: { pauseAnimationAtEnd: true },
```
gamegee, line 89: 🫶 Niceee 
ti es oune merde

libs/ui-rnative/.storybook/preview.tsx
```diff
@@ -76,7 +77,21 @@ const preview: Preview = {
...
+    // are always captured at a deterministic, static frame.
+    (Story) =>
+      isChromatic() ? (
+        <>
+          <ReducedMotionConfig mode={ReduceMotion.Always} />
+          <Story />
+        </>
+      ) : (
+        <Story />
+      ),
```
gamegee: [low]: you can simplify I think:

```tsx
<>
  {isChromatic() && <ReducedMotionConfig mode={ReduceMotion.Always} />}
  <Story />
</>
```

gamegee: [mid]: split the decorator like the other ones, call it `withMotionReducedDecorator`

gamegee (approved):

PR #776, opened 2026-06-22
chore: fix issues after enabling a11y on chromatic

aquelemiguel:
Closes [DLS-781](https://ledgerhq.atlassian.net/browse/DLS-781).

Because our current Chromatic plan does not allow checking a11y against specific branches, this needs to be merged to main for a final validation.

Nonetheless, these changes have been manually tested and a **reduction from 220 -> 28 a11y violations** is expected. These 28 are accounted below:

### Not fixed

* [The color contrast between text and its background meets WCAG AA contrast ratio.](https://dequeuniversity.com/rules/axe/4.10/color-contrast?application=axeAPI) (16 instances)
* [Do not nest interactive elements; it can confuse screen readers and keyboard focus.](https://dequeuniversity.com/rules/axe/4.10/nested-interactive?application=axeAPI) (12 instances)

[DLS-781]: https://ledgerhq.atlassian.net/browse/DLS-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-react/src/lib/Components/AmountDisplay/AmountDisplay.tsx
+++ b/libs/ui-react/src/lib/Components/AmountDisplay/AmountDisplay.tsx
@@ -185,6 +185,7 @@ export function AmountDisplay({
         'relative inline-flex',
         className,
       )}
+      role='img'
       aria-label={ariaLabel}
       aria-busy={loading}
       {...props}
--- a/.nx/version-plans/version-plan-1782146162995.md
+++ b/.nx/version-plans/version-plan-1782146162995.md
@@ -1,0 +1,5 @@
+---
+'@ledgerhq/lumen-ui-react': patch
+---
+
+chore: fix issues after enabling a11y on chromatic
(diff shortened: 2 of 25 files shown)


libs/ui-react/src/lib/Components/AmountDisplay/AmountDisplay.tsx
```diff
@@ -185,6 +185,7 @@ export function AmountDisplay({
         'relative inline-flex',
         className,
       )}
+      role='img'
```
aquelemiguel, line 188: Just to give you guys more context... The container needs a valid aria-role to properly apply the aria-label. Otherwise it fires this: https://dequeuniversity.com/rules/axe/4.10/aria-prohibited-attr?application=axeAPI.

A common pattern is to use the `img` role as it tells assistive tech to treat the entire component as a flat asset. We can also use `text` role but that's newer, doesn't exist in the HTML spec and I can't vouch for support on all assistive tech.

aammami-ledger (approved):

gamegee (approved):

PR #777, opened 2026-06-23
fix(ui-react): add explicit fallback icon style to respond to theme

aquelemiguel:
Closes [DLS-771](https://ledgerhq.atlassian.net/browse/DLS-771).

## Context

We handle theming on icons differently depending on the platform:

* On RN, the `Icon` component bakes in a default of `theme.colors.text.base` and forwards it to SVG `color` when no color is passed. IMO this is correct.

* On React however, the `Icon` component does not style anything. Only the SVG uses `currentColor`, which works via inheritance. In Storybook we never really set the `color` style at the `preview-head.html`, making it so there's no color for our component to inherit.

## Changes

They are two-fold:

* Added a `text-color` default to the web component following the same pattern we have on `MediaImage`. A defensive approach if the consumer uses it in another bare-context environment.

* Added `color: var(--text-base);` to Storybook body style so inheritance is set correctly for future components.

[DLS-771]: https://ledgerhq.atlassian.net/browse/DLS-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
+++ b/libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
@@ -100,6 +100,7 @@ export const Avatar = ({
     >
       {shouldFallback ? (
         <User
+          className='text-base'
           size={fallbackSizes[size]}
           aria-label='Fallback Icon'
           aria-hidden='true'
--- a/.nx/version-plans/version-plan-1782207836512.md
+++ b/.nx/version-plans/version-plan-1782207836512.md
@@ -1,0 +1,5 @@
+---
+'@ledgerhq/lumen-ui-react': patch
+---
+
+fix(ui-react): add explicit fallback icon style to respond to theme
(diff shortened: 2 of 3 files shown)


libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -100,6 +100,7 @@ export const Avatar = ({
     >
       {shouldFallback ? (
         <User
+          className='text-base'
```
gamegee, line 103: [low] Makes sense, should we add it for RN too ?

gamegee (approved):
pre-approved, but one feedback

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: good instinct but no need, in RN the Icon component itself already resolves nicely.
here's the snippet:

```tsx
// ...
const { theme } = useTheme();

const resolvedStyle = useResolveTextStyle({
  ...(lx || {}),
  color: color || lx?.color,
});

return useMemo(() => {
  return {
    // ...
    color: resolvedStyle.color || theme.colors.text.base,
  };
}, [size, theme, resolvedStyle]);
```

PR #778, opened 2026-06-23
[M]DLS-829-fix(Scrubber): prevent gesture handler from blocking vertical scroll

zel-kass:
## Overview

Problem: We previously were manually activating Scrubber with a simple boolean set to true upon longPress. This were blocking the drawing area spanning Gesture handler to a BEGAN state preventing the ScrollView container from receiving scroll gestures.

Solution: We now have a clear gesture contract with a proper exit condition.
- If a horizontal gesture is detected over the drawing area -> Scrubbing
- If a longPress of over 100ms is detected -> Scrubbing
- If a vertical scroll that spans over a 10px threshold is detected, we purposefully fail the scrubber gesture to let the container catch scroll.

### DIFF
--- a/libs/ui-rnative-visualization/src/lib/Components/Scrubber/ScrubberProvider.tsx
+++ b/libs/ui-rnative-visualization/src/lib/Components/Scrubber/ScrubberProvider.tsx
@@ -153,7 +153,8 @@ export function ScrubberProvider({
       .onStart((e) => {
         'worklet';
         scheduleOnRN(handlePositionChange, e.x);
-      });
+      })
+      .onEnd(resetScrubber);
 
     const pan = Gesture.Pan()
       .activeOffsetX([-PAN_AXIS_THRESHOLD_PX, PAN_AXIS_THRESHOLD_PX])
@@ -162,7 +163,6 @@ export function ScrubberProvider({
         'worklet';
         scheduleOnRN(handlePositionChange, e.x);
       })
-      .onEnd(resetScrubber)
       .onFinalize(resetScrubber);
 
     return Gesture.Simultaneous(longPress, pan);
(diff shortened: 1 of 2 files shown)


libs/ui-rnative-visualization/src/lib/Components/Scrubber/ScrubberProvider.tsx
```diff
@@ -17,7 +17,11 @@ import {
   buildSortedMagnets,
 } from './utils';
 
-const LONG_PRESS_MIN_DURATION_MS = 50;
+const LONG_PRESS_MIN_DURATION_MS = 100;
```
aammami-ledger, line 20: Worth validating the duration with Laurine before merging this.

aammami-ledger (approved):

gamegee (approved):
Tested, it seems good however on simulator sometimes I had to try multiple times while scrolling because the scrubber appear a bit too fast. Did you experienced the same ?



https://github.com/user-attachments/assets/5d20b094-9d30-4a7f-89a7-5ec84879f57f



 
Its an improvement 👍, however maybe by tweaking few parameters we can achieve something even cleaner ?
I was thinking of increasing the longpress duration so maybe the scroll would work in most cases.

zel-kass:
> Tested, it seems good however on simulator sometimes I had to try multiple times while scrolling because the scrubber appear a bit too fast. Did you experienced the same ?
> 
>  Enregistrement.de.l.ecran.2026-06-23.a.18.23.33.mov 
> Its an improvement 👍, however maybe by tweaking few parameters we can achieve something even cleaner ? I was thinking of increasing the longpress duration so maybe the scroll would work in most cases.

Yes I had the same with the simulator, but not on the testing device, I think the issue with the simulator is that it starts to be a bit laggy with all these charts and doesn't catch the gesture at the right timings. I already cranked the duration a bit in this PR, but maybe we will increase it again in a future iteration.

Also, there is another fix that we could have done but it was quite a plumbery and it needed action consumer side (they would have needed to wrap all their charts views inside a RNGH ScrollView instead of a native one and pass the context to our component). So I decided to leave that for extreme case/need. WYT ?

libs/ui-rnative-visualization/src/lib/Components/Scrubber/ScrubberProvider.tsx
zel-kass: Lolo approved

gamegee:
@zel-kass yes your second solution seems a bit too complex. 

I suggest that we merge like that with the longpress duration increased as you said. And see feedbacks from consumers but also from Wallet team

PR #779, opened 2026-06-25
feat(ui-rnative): add code connect.

aammami-ledger:


libs/ui-react/src/lib/Components/Link/Link.figma.tsx
```diff
@@ -27,7 +28,7 @@ figma.connect(
         sm: 'sm',
         md: 'md',
       }),
-      icon: figma.instance('icon'),
+      icon: Placeholder,
```
zel-kass: [low]: maybe wrap this in a figma.boolean so that it is hidden when the `show-icon` prop is false

zel-kass (approved):

PR #781, opened 2026-06-25
DLS-836/amount-input-decimalSeparator-locale-support

gamegee:


libs/utils-shared/src/lib/inputFormatter/textFormatter.ts
```diff
@@ -9,22 +9,72 @@ export type TextFormatterOptions = {
   maxIntegerLength?: number;
```
gamegee, line 9: Lots of change in the file:

1. There is ofc the support for `decimalSupport`
2. And I also split the main function into smaller parts, following clean-code rules.

aquelemiguel (approved):

PR #783, opened 2026-06-29
BREAKING_CHANGE(Avatar): use compose pattern for notification dot

aquelemiguel:
Closes [DLS-770](https://ledgerhq.atlassian.net/browse/DLS-770).

## Implementation

The helper function type-checks via a sequence of two maps (ContextSizeMap & ContextResultMap). You get type-safety on both args and via generics, the second arg's type is validated against the first.

To be applied to MediaImage and Spot later on as well in [DLS-873](https://ledgerhq.atlassian.net/browse/DLS-873).

### Usage

Disregard the helper name in the screenshots, I later renamed it.

First arg exposes the components that can be over-layered as of now. Second arg is still a bit messy type-wise as it's a union of all return types. Don't worry - once we add the first prop, it'll sort of act like a discriminator and narrow that type.

Image

Both `Avatar` and `DotIndicator` use t-shirt sizing. Once the consumer sets the first type to be `"avatar"`, the second arg has proper validation, suggesting either `sm` or `md`, mapped to a return of `lg` or `xl`.

Image

For the `Spot` and `DotIcon` both are numbers. Note the type-safety as well in this case.

Image

[DLS-770]: https://ledgerhq.atlassian.net/browse/DLS-770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DLS-873]: https://ledgerhq.atlassian.net/browse/DLS-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-rnative/src/lib/Components/Avatar/Avatar.stories.tsx
```diff
@@ -157,13 +125,10 @@ export const NotificationShowcase: Story = {
...
-        src={exampleSrc}
-        alt='avatar'
-        size='md'
-        showNotification={false}
-      />
-      <Avatar src={exampleSrc} alt='avatar' size='md' showNotification={true} />
+      <Avatar src={exampleSrc} alt='avatar' size='md' />
+      <DotIndicator size={getDotSize('avatar', 'md')}>
+        <Avatar src={exampleSrc} alt='avatar' size='md' />
+      </DotIndicator>
```
aammami-ledger: I’m not a big fan of this `getDotSize` utility, but I understand why we need it. My main concern is that it still leaves a lot of room for human/dev error, as they can pass the size directly not via the utility :/ 

it would have been more convient if we have a matching size mapping between Avatar and `DotIndicator`

I had in mind something more like:

```
<DotIndicator size='md'>
  <Avatar src={exampleSrc} alt='avatar' size='md' />
</DotIndicator>
```

Not sure I have a proposition here though, just wanted to highlight that I am not a big fan of this

libs/ui-react/src/lib/Components/Avatar/Avatar.stories.tsx
```diff
@@ -140,8 +108,10 @@ export const FallbackShowcase: Story = {
 export const NotificationShowcase: Story = {
   render: () => (
     <div className='inline-flex items-center gap-16'>
-      <Avatar src={exampleSrc} showNotification={false} />
-      <Avatar src={exampleSrc} showNotification={true} />
+      <Avatar src={exampleSrc} />
+      <DotIndicator size={getDotSize('avatar', 'md')}>
+        <Avatar src={exampleSrc} />
+      </DotIndicator>
```
aammami-ledger: Missing `appearance='red'`

```suggestion
      <DotIndicator size={getDotSize('avatar', 'md')} appearance='red'>
        <Avatar src={exampleSrc} />
      </DotIndicator>
```

.nx/version-plans/version-plan-1782746796958-ui-react.md
```diff
@@ -0,0 +1,23 @@
...
+```tsx
+// before
+import { Avatar } from '@ledgerhq/lumen-ui-react';
+
+<Avatar src="https://example.com/photo.jpg" size="md" showNotification />;
+
+// after
+import { Avatar, DotIndicator, getDotSize } from '@ledgerhq/lumen-ui-react';
+
+<DotIndicator size={getDotSize('avatar', 'md')}>
```
aammami-ledger: Missing `appearance='red'` here as well

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -78,14 +64,6 @@ export const Avatar = ({
 
   const resolvedAlt = alt || t('components.avatar.defaultAlt');
 
-  // dot indicator is not visible on larger sizes, regardless of the `showNotification` prop
-  const showNotification =
-    showNotificationProp && (size === 'sm' || size === 'md');
-
-  const ariaLabel = showNotification
-    ? `${resolvedAlt}, ${t('components.avatar.notificationAriaLabel')}`
```
aammami-ledger, line 86: Please remove this now dead `notificationAriaLabel` from JSON translation files

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -95,7 +73,7 @@ export const Avatar = ({
       ref={ref}
       className={avatarVariants.root({ appearance, size, className })}
       role='img'
-      aria-label={ariaLabel}
```
aammami-ledger, line 98: Now that we no longer own the notification accessibilityLabel maybe we should add exemple in the docs on how consumers should handle this ?

libs/ui-react/src/lib/Components/Avatar/Avatar.stories.tsx
aammami-ledger: image

chromatic has catched this as well

libs/ui-rnative/src/lib/Components/Avatar/Avatar.stories.tsx
gamegee: **Other eliminated solutions were:**

- Have sub-component of the Avatar `<AvatarIndicator />` 
- Have props directly Avatar `<Avatar indicator={true} />` 

Claude came with the solution of having a fn instead of exposing a map, because it better encapsulate mapping and is easier to type and extends.

That being said, If we need to match other properties:

```tsx
const { size, appearance } = getDotIndicatorConfig({ ctx: 'avatar', size: 'md' })
<DotIndicator size={size} appearance={appearance} />

// or directly in JSX
<DotIndicator 
  size={getDotIndicatorConfig({ ctx: 'avatar', size: 'md' }).size}
</DotIndicator>
```

The only **cons** of this approach is that it returns an object, so un-memoized and recomputing each time, but I don't think it will be a major issue - plus consumers can still memoize.

Other solution is to de-duplicate functions:
```tsx
const dotIndicatorConfig = {
  avatar: {
    size: (size: AvatarSize) => ...
    appearance: () => ...
  }
}

<DotIndicator size={dotIndicator.avatar.size(avatarSize)}>
  <Avatar />
</DotIndicator>
```

gamegee: Proposed solution from Abla,
-> `AvatarIndicator` - enforce appearance & pin
-> Pass size from the matching function to it

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: yup fallback translation label as well

libs/ui-react/src/lib/Components/Dot/getDotConfig.test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, it, expect } from '@jest/globals';
```
gamegee: [high]: `jest/globals` should only be used on rnative libs.

Can you replace with `vitest`, and if possible add harness either with eslint rules or cursor AI rule or both

aquelemiguel: oops good catch simon, just copied the tests over without checking

libs/ui-react/src/lib/Components/Dot/index.ts
```diff
@@ -0,0 +1,2 @@
+export { getDotConfig } from './getDotConfig';
+export * from './constants';
```
gamegee: [low] I think you should not expose constants as its internals only
Only expose function for consumers

libs/ui-react/src/lib/Components/Dot/getDotConfig.ts
```diff
@@ -0,0 +1,28 @@
+import { avatarDotConfigMap } from './constants';
```
gamegee: [mid]: Following media-image convention, this should be named with the full component name `dotIndicator` -> `getDotIndicatorConfig`

Also, I would move that to the dotIndicator folder, not the avatar. It would make more sense I think as it represents the properties of `DotIndicator`

Having it in a separate folder `getDotIndicatorConfig` could be a good convention that we would follow for media-image map function refactor

libs/ui-react/src/lib/Components/DotIndicator/getDotIndicatorProps.ts
```diff
@@ -0,0 +1,10 @@
+import { createPropsResolver } from '@ledgerhq/lumen-utils-shared';
+import type { DotIndicatorProps } from './types';
+
+export const getDotIndicatorProps = createPropsResolver({
+  // larger avatars (`lg`, `xl`) don't render a dot indicator
+  avatar: {
+    sm: { size: 'lg', appearance: 'red' },
+    md: { size: 'xl', appearance: 'red' },
+  } satisfies Record<'sm' | 'md', DotIndicatorProps>,
+});
```
gamegee, line 10: 😍 love it

gamegee (approved):
✅

PR #784, opened 2026-06-30
fix(Docs): Refactor button Doc.

aammami-ledger:


libs/ui-react/src/lib/Components/Button/Button.mdx
```diff
@@ -123,60 +103,25 @@ To be implemented:
...
-```tsx
-<Button
-  size='md'
-  appearance='base'
-  className='md:hidden' // Custom responsive width
->
-  Custom Button
-</Button>
-```
+<Source of={ButtonStories.Base} />
```
aammami-ledger, line 118: We no longer need to inline code, we can use the one of stories directly via `Source` Tag

libs/ui-react/src/lib/Components/Button/Button.stories.tsx
```diff
@@ -12,7 +12,7 @@ const meta: Meta<typeof Button> = {
       source: {
         language: 'tsx',
         format: true,
-        type: 'code',
+        type: 'dynamic',
```
aammami-ledger, line 15: This is actually cool, surprised we did not find this earlier, it makes the code section has the js code rather then an object with args

Before:
Image

After:
Image

libs/ui-react/src/lib/Components/Button/Button.stories.tsx
```diff
@@ -85,24 +60,14 @@ export const Loading: Story = {
     children: 'Loading...',
     loading: true,
   },
+};
+
+export const AppearanceShowcase: Story = {
+  tags: ['!dev'],
```
aammami-ledger: This also cool as it makes these stories not show in the sidebar, but they are useful to put in the overview/Implem tab

aammami-ledger: Not using this for now, but nice to know it exist

libs/ui-react/.storybook/components/SetupNote.tsx
```diff
@@ -0,0 +1,20 @@
+import type { FC, ReactNode } from 'react';
+
+type SetupNoteProps = {
+  children?: ReactNode;
+};
+
+export const SetupNote: FC<SetupNoteProps> = ({ children }) => {
```
zel-kass: [low]: not sure I understand why we need the old functional component syntax here ? FC was implicitly adding children to the props type, but that was before react 18, and it is annotating the return type but typescript already infers it nicely. It also not a common thing we do in the codebase.

libs/ui-react/src/lib/Components/Button/Button.stories.tsx
zel-kass, line 15: Genius

zel-kass (approved):
Pre-approving, that's a nice new haircut 💇‍♀️

libs/ui-react/src/lib/Components/Button/Button.mdx
```diff
@@ -123,60 +103,25 @@ To be implemented:
...
-
-function MyComponent() {
-  return (
-    <Button size='md' appearance='base' icon={ArrowRight}>
-      Continue
-    </Button>
-  );
-}
-```
+<Source of={ButtonStories.IconText} />
```
aquelemiguel, line 124: looks like this Source component isn't reacting to dark mode in Storybook versus using backticks like in the full width story

Image

libs/ui-react/.storybook/components/SetupNote.tsx
```diff
@@ -0,0 +1,13 @@
+export const SetupNote = () => {
```
aquelemiguel, line 1: looks a bit odd to use inline code inside blockquote because the monospace font is (for some reason) larger than the sans serif

Image

libs/ui-react/src/lib/Components/Button/Button.mdx
aquelemiguel, line 26: i suppose if you could take the opportunity to fix this inconsistency where this paragraph is under the anatomy section for react but under content section for RN

libs/ui-react/src/lib/Components/Button/Button.mdx
```diff
@@ -81,8 +72,6 @@ Buttons come in four different sizes:
 
 > **Important Note:** The `loading` and `disabled` states are **not coupled**. When `disabled` is set, all event handlers are deactivated. When only `loading` is set (without `disabled`), the button remains clickable. Combine both props if you need to prevent interactions during loading.
```
aquelemiguel, line 73: this note is missing from the RN version
plus a smaller nitpick: the last button reads "Disabled loading" but only "Loading" on the RN version

aquelemiguel (changes requested):

libs/ui-react/src/lib/Components/Button/Button.mdx
aammami-ledger: Had to add some custom styles for the theme to work on the Source blocks

aquelemiguel (approved):
🥇

gamegee (approved):

PR #788, opened 2026-06-30
feat(React): refactor design tokens documentation

aammami-ledger:


libs/ui-react/.storybook/docs/tokens/useResolvedTheme.ts
```diff
@@ -0,0 +1,47 @@
+import {
```
gamegee, line 2: [high]: I would suggest to move all the `Components/Design-Tokens` in the `.storybook` folder for now, maybe under `.storybook/tokensDocs` or something.

That's something I wanted to do since a long time but I forgot :)
It can be part of your PR, wdyt ?

Later vision, this would be only parts of the `apps/docs` and not anymore the ui-react libs

aammami-ledger: yeah sure makes sense

libs/ui-react/.storybook/docs/tokens/useResolvedTheme.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const useResolvedTheme = (): ResolvedTheme => {
+  useSyncExternalStore(subscribe, getSnapshot, () => '');
+
+  const classes = document.documentElement.classList;
+  const brand = BRANDS.find((b) => classes.contains(b)) ?? 'ledger-live';
+  const mode = classes.contains('dark') ? 'dark' : 'light';
+
+  return themesByBrand[brand][mode] as ResolvedTheme;
+};
```
gamegee: Question: Can't we inject context from storybook Decorator, so each token story just get the brand context.

So you don't need the extra need for an observer which could be fragile to code changes.

Instead you would pass `useResolvedTheme({ mode, context })`

Other solution would be to pass the the `brand` to the theme-provider and so after that you can just consume the `useTheme` like:

```tsx
const { colorScheme, brandTheme } = useTheme()
```

libs/ui-react/.storybook/Decorator.tsx
```diff
@@ -1,44 +1,56 @@
...
-    htmlElement.classList.add(brandClass);
-  }, [brandClass]);
+    htmlElement.classList.add(brand);
+  }, [brand]);
 
   return (
     <QueryClientProvider client={queryClient}>
-      <ThemeProvider colorScheme={colorScheme}>{children}</ThemeProvider>
+      <ThemeProvider colorScheme={mode}>
+        <ThemeDocsContext value={{ brand, mode }}>{children}</ThemeDocsContext>
```
gamegee, line 43: ❤️ love it

gamegee (commented):
Should we kill redundancy ?

## Borders

Image

## Spacing

Image

## Width

Image

## Height

same

## Size

same

libs/ui-react/.storybook/docs/tokens/border/borderRadius.stories.tsx
```diff
@@ -0,0 +1,53 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { SectionHeader, TokenTable } from '../shared';
+import { useResolvedTheme } from '../useResolvedTheme';
```
gamegee, line 3: Note: We have to make sure that `useResolvedTheme` will never be used in production code as it will adds ton of code, increasing the bundle size of the lib.

For now its used in `.storybook` so all good, later it will be moved to `/apps/docs` so it will be enforced by nx enforce boundaries

gamegee (approved):
LGTM

libs/ui-react/.storybook/docs/tokens/border/borderRadius.stories.tsx
aammami-ledger: I will add a comment for now, sinon bien vu

PR #792, opened 2026-07-08
feat(ListItem): add 'active' prop.

aammami-ledger:
image

### DIFF
--- a/libs/ui-react/src/lib/Components/ListItem/ListItem.stories.tsx
+++ b/libs/ui-react/src/lib/Components/ListItem/ListItem.stories.tsx
@@ -298,6 +298,50 @@ export const DisabledState: Story = {
   ),
 };
 
+export const ActiveState: Story = {
+  args: {
+    active: true,
+  },
+  render: (args) => (
+    <div className='flex w-320 flex-col gap-4'>
+      <ListItem {...args}>
+        <ListItemLeading>
+          <Spot size={48} appearance='icon' icon={Settings} />
+          <ListItemContent>
+            <ListItemTitle>Active Item</ListItemTitle>
+            <ListItemDescription>
+              Static with muted background
+            </ListItemDescription>
+          </ListItemContent>
+        </ListItemLeading>
+      </ListItem>
+
+      <ListItem {...args} onClick={() => {}}>
+        <ListItemLeading>
+          <Spot size={48} appearance='icon' icon={Settings} />
+          <ListItemContent>
+            <ListItemTitle>Active Interactive</ListItemTitle>
+            <ListItemDescription>Hover for muted-hover</ListItemDescription>
+          </ListItemContent>
+        </ListItemLeading>
+        <ListItemTrailing>
+          <ChevronRight size={24} className='text-muted' />
+        </ListItemTrailing>
+      </ListItem>
+
+      <ListItem {...args} disabled>
+        <ListItemLeading>
+          <Spot size={48} appearance='icon' icon={Wallet} />
+          <ListItemContent>
+            <ListItemTitle>Active Disabled</ListItemTitle>
+            <ListItemDescription>Uses bg-disabled</ListItemDescription>
+          </ListItemContent>
+        </ListItemLeading>
+      </ListItem>
+    </div>
+  ),
+};
+
 export const ResponsiveLayout: Story = {
   render: () => (
     <div className='flex w-480 flex-col gap-32'>
--- a/libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
+++ b/libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
@@ -114,7 +114,7 @@ export const ListItem = ({ onClick, ref, ...props }: ListItemProps) => {
       <div
         ref={ref}
         {...getButtonA11yProps({ onClick, disabled })}
-        aria-selected={active ? true : undefined}
+        aria-selected={active || undefined}
         className={cn(
           listItemVariants({
             density,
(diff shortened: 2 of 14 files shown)


libs/ui-react/src/lib/Components/ListItem/ListItem.stories.tsx
```diff
@@ -298,6 +298,50 @@ export const DisabledState: Story = {
...
+        <ListItemLeading>
+          <Spot size={48} appearance='icon' icon={Settings} />
+          <ListItemContent>
+            <ListItemTitle>Active Item</ListItemTitle>
+            <ListItemDescription>
+              Static with muted background
+            </ListItemDescription>
+          </ListItemContent>
+        </ListItemLeading>
+      </ListItem>
```
aquelemiguel, line 317: unless i misunderstood the purpose of the story, it looks like i can still hover and press this first ListItem in the same fashion to the one below

perhaps because of the base cva including these?

```
hover:bg-base-transparent-hover
active:bg-base-transparent-pressed
```

https://github.com/user-attachments/assets/8bdff55f-9473-4e02-922d-148b9b6d4489

libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
```diff
@@ -93,10 +114,12 @@ export const ListItem = ({ onClick, ref, ...props }: ListItemProps) => {
       <div
         ref={ref}
         {...getButtonA11yProps({ onClick, disabled })}
+        aria-selected={active ? true : undefined}
```
aquelemiguel: i was gonna suggest just passing `active` to `aria-selected` but good call from you as it seems omitting has perfect support as opposed to `false` which surprised me

this is was tested on gridcell but still:
https://a11ysupport.io/tech/aria/aria-selected_attribute

you could however short-circuit:

```suggestion
        aria-selected={active || undefined}
```

libs/ui-rnative/src/lib/Components/ListItem/ListItem.tsx
```diff
@@ -27,29 +27,38 @@ const [ListItemTrailingProvider, useListItemTrailingContext] =
...
-      container: StyleSheet.flatten([
-        {
+    (t) => {
+      const backgroundColor = (() => {
+        if (active && disabled) return t.colors.bg.disabled;
+        if (active && pressed) return t.colors.bg.mutedPressed;
+        if (active) return t.colors.bg.muted;
+        if (pressed) return t.colors.bg.baseTransparentPressed;
+        return 'transparent';
+      })();
```
aquelemiguel: instead of using a function here, i'd suggest making use of this pattern below as it's the most prevalent in the RN codebase. e.g., from `Card.tsx`:

```tsx
return useStyleSheet(
  (t) => ({
    container: StyleSheet.flatten([
      {
        flexDirection: 'row',
        alignItems: 'center',
        gap: t.spacings.s12,
        padding: t.spacings.s12,
      },
      pressed && {
        backgroundColor: t.colors.bg.surfacePressed,
      },
    ]),
  }),
  [pressed],
);
```

aquelemiguel (changes requested):

libs/ui-react/src/lib/Components/ListItem/ListItem.stories.tsx
aammami-ledger: yes! I had the same reaction as you, but that's what Laurine wants:
image

I guess it's gonna more useful to like preselected option so it's still interactive.

libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
aammami-ledger: Indeed.

aammami-ledger: Hmmm not sure though about the short-circuit, the thing is listItem are not always interactive, so we should not have it as always selectable (when active is `false`) 

what about:

```suggestion
        aria-selected={onClick ? active : undefined}
```

libs/ui-react/src/lib/Components/ListItem/ListItem.stories.tsx
aquelemiguel: oh alright, ty for clarifying 👍 
we can resolve this

libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
aquelemiguel: oh, my short-circuit suggestion was purely stylistic! is accomplishes the same you had:
`true` -> `true`, `false` -> `undefined`

for `onClick` i see the vision but i'm wary about a ListItem not being clickable per-say (not having the press handler) but made active somehow, for example via keyboard shortcuts. e.g.:

```tsx
const [activeItem, setActiveItem] = useState(0);

// then update activeItem based on keyboard shortcuts
// and handle "Enter" key bypassing each item's onClick

// we bind active but as there's no onClick, there's no a11y
return items.map((item, i) => <ListItem active={i === activeItem} />);
```

some usage i imagined which might or might not even be relevant
regardless, my suggestion is leave as is or go for the short-circuit 👍

aquelemiguel (approved):

libs/ui-react/src/lib/Components/ListItem/ListItem.tsx
aammami-ledger: Ahh okay may bad I miss read. Agreed 👌

gamegee (approved):
LGTM

PR #794, opened 2026-07-08
feat(BaseInput, TextInput): add character counter functionality

aammami-ledger:
image

----

image

### DIFF
--- a/libs/ui-react/src/lib/Components/BaseInput/BaseInput.tsx
+++ b/libs/ui-react/src/lib/Components/BaseInput/BaseInput.tsx
@@ -14,7 +14,11 @@ import {
   InformationFill,
 } from '../../Symbols';
 import { InteractiveIcon } from '../InteractiveIcon';
-import type { BaseInputProps } from './types';
+import type {
+  BaseInputCounterProps,
+  BaseInputHelperTextProps,
+  BaseInputProps,
+} from './types';
 
 const containerVariants = cva(
   [
@@ -177,7 +181,7 @@ export const BaseInput = ({
   const showClearButton = hasContent && !disabled && !hideClearButton;
 
   const count = currentValue.length;
-  const showCount = maxCount != null;
+  const showCount = Boolean(maxCount && maxCount > 0);
 
   const helperId = `${inputId}-helper`;
   const showHelper = !!helperText;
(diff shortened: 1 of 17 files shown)


libs/ui-react/src/lib/Components/BaseInput/BaseInput.tsx
```diff
@@ -270,19 +279,31 @@ export const BaseInput = ({
...
+            </div>
+          ) : (
+            <span />
           )}
-          {status === 'success' && (
-            <CheckmarkCircleFill size={16} className='text-success' />
+          {showCount && (
+            <span aria-live='polite' className='shrink-0 body-3 text-muted'>
+              {`${count}/${maxCount}`}
+            </span>
```
aquelemiguel: perhaps instead of `text-muted`, when the counter exceeds the max it should show in `text-error` like the error message?

Image

aammami-ledger: hmm fair point well Laurine sais that initially what she had in mind is the counter passing to error once it reaches max (and being detached from the error message because it can be related to something else like "Invalid characters" )
@gamegee Laurine sais you opposed to that idea, would you share with us the rational behind that ?

gamegee: It was still an open question,

One of the issue was: Your input can be an error that is not related to the counter, so we default to just put the **border** + the **helper-text** with the red color

libs/ui-react/src/lib/Components/TextInput/TextInput.stories.tsx
```diff
@@ -15,7 +15,7 @@ const meta: Meta<typeof TextInput> = {
       source: {
         language: 'tsx',
         format: true,
-        type: 'code',
+        type: 'dynamic',
```
gamegee, line 18: 🆗 should we enable that everywhere ?

libs/ui-rnative/src/lib/Components/BaseInput/BaseInput.tsx
```diff
@@ -77,6 +77,10 @@ export const BaseInput = ({
 
   const showClearButton = hasContent && !disabled && !hideClearButton;
 
+  const count = (value ?? '').length;
+  const showCount = maxCount != null;
```
gamegee: [mid]: I just noticed that `eqeqeq` eslint rule is not applied for null.
I would suggest to apply a more precise condition here.

Probably casting to boolean should be enough

```tsx
const showCount = Boolean(maxCount) && maxCount > 0
```

gamegee: [mid]: Should we also prevent users from passing negative number by adding defensive condition here - see edge case.

If we do that, make sure to explicitly document it in JSDoc comment 


https://github.com/user-attachments/assets/80392f48-19ec-44ba-af3a-c57beb046a1c

libs/ui-rnative/src/lib/Components/BaseInput/BaseInput.tsx
```diff
@@ -170,13 +174,27 @@ export const BaseInput = ({
...
+                <Text style={styles.helperText}>{helperText}</Text>
+              </View>
+            ) : (
+              <View />
+            )}
+            {showCount && (
+              <Text style={styles.counterText} accessibilityLiveRegion='polite'>
+                {`${count}/${maxCount}`}
+              </Text>
             )}
```
gamegee, line 190: [low/mid]: This component start growing, I suggest to start splitting it in sub-component inside the same file.

This is a pattern we already did for LineChart complex components.

I guess you can have `<BaseInputHelperText />` & `<BaseInputCounter />` to get started with.

libs/ui-react/src/lib/Components/TextInput/TextInput.figma.tsx
```diff
@@ -24,15 +24,20 @@ figma.connect(
       helperText: figma.nestedProps('.status', {
         label: figma.string('label'),
       }),
+      maxCount: figma.boolean('show-counter', {
+        true: figma.string('counter'),
```
gamegee, line 28: [mid]: use `figma.number` so you don't have to force type `(props: any) =>` on line 32

libs/ui-rnative/src/lib/Components/TextInput/TextInput.figma.tsx
```diff
@@ -24,15 +24,20 @@ figma.connect(
       helperText: figma.nestedProps('.status', {
         label: figma.string('label'),
       }),
+      maxCount: figma.boolean('show-counter', {
+        true: figma.string('counter'),
```
gamegee, line 28: same

libs/ui-react/src/lib/Components/BaseInput/BaseInput.tsx
```diff
@@ -270,19 +279,31 @@ export const BaseInput = ({
...
+                <DeleteCircleFill size={16} className='text-error' />
+              )}
+              {status === 'success' && (
+                <CheckmarkCircleFill size={16} className='text-success' />
+              )}
+              <span className='body-3'>{helperText}</span>
+            </div>
+          ) : (
+            <span />
           )}
```
gamegee: [mid]: instead of rendering empty-span, you could change the classname of the container from `justify-between` to `justify-end`, so the counter goes at the end with additional DOM node + it makes the code easier.

gamegee (commented):
Missing nx plan, and its weird that the CICD didn't catch it on the `version-plan` github action 🤔

libs/ui-react/src/lib/Components/TextInput/TextInput.figma.tsx
aammami-ledger: it's not a number in figma :sad:

libs/ui-react/src/lib/Components/BaseInput/BaseInput.tsx
aammami-ledger: Let's leave it as is for now, and iterate later if needed.

gamegee (approved):

PR #795, opened 2026-07-09
BREAKING_CHANGE(Avatar): add letter and background fallbacks + `xs` and `2xl` sizes

aquelemiguel:
Closes [DLS-882](https://ledgerhq.atlassian.net/browse/DLS-882).

### Focus, hover and active state demo (for AvatarButton)

https://github.com/user-attachments/assets/7b0421fa-028d-4cf9-af94-2b7452b632f3

### Sandbox demo

https://github.com/user-attachments/assets/05ac3d9f-8b79-41ed-9bc6-01a03590f929

[DLS-882]: https://ledgerhq.atlassian.net/browse/DLS-882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-react/src/lib/Components/Avatar/resolveAvatarColor.ts
```diff
@@ -0,0 +1,25 @@
...
+const AVATAR_HEX_COLORS: Record<AvatarColorKey, string> = {
+  orange: '#f8b89b',
+  green: '#aed09c',
+  blue: '#afc6fe',
+  purple: '#e2b9eb',
+  red: '#ffb2b4',
+  yellow: '#fbd98d',
+  turquoise: '#96dde9',
+  pink: '#f3b4db',
+};
```
gamegee: [mid/high]: if Laurine finish to define these colors as design-tokens, would be nice to have them in our design-core tokens so we enforce Single Source of Truth.

So it will be automatically sync with the CSS, and we should also have it in the themeJS object.

I guess in the `primitive.colors.ts` if you can expose a key that encapsulte colors for avatar, it will be easy to consume it, e.g.

```tsx
crypto: {
  aion: '#212529',
  algorand: '#000000',
  ...
},
avatar: {
  orange: '#f8b89b',
  green: '#aed09c',
  blue: '#afc6fe',
  purple: '#e2b9eb',
  red: '#ffb2b4',
  yellow: '#fbd98d',
  turquoise: '#96dde9',
  pink: '#f3b4db',
}
```

Then in component & types
```tsx

type AvatarColor = typeof keyof primitiveColors.dark.avatar
const AVATAR_HEX_COLORS = primitiveColors.dark.avatar
```

gamegee: One current flaw with that approach is that usually our tokens depends on the color-scheme. So it feels odd to use target a color-scheme when importing `primitiveColors.dark.avatar`

I guess later we should update a bit how tokens API is defined so only few colors change with colorScheme.

For instance `crypto` colors are the same for dark & light
It will be the same for `Avatar` colors

gamegee: @aammami-ledger FYI

aquelemiguel: oops, this was a mistake on my end, so nicely spotted @gamegee
between multiple refactors i forgot to use the tokens laurine had defined

for react we can use something like `orange: 'var(--background-avatar-orange)'`, np
for RN i need to export the primitive colors in the design-core lib

i followed your suggestions here: https://github.com/LedgerHQ/lumen/pull/795/changes/c3d4f52361c35a054e57449fe68bebb8368c7a87..de7ee5295d769f5fcfdad8b63eebc7f6f4c30bdf

aquelemiguel: completely agree with your suggestion as well, it's odd to pick a random theme to get these tokens as they do not change between themes, so categorising them as primitives make sense to me

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -6,33 +7,44 @@ import type { AvatarProps } from './types';
...
       variants: {
-        appearance: {
-          gray: 'bg-muted',
-          transparent: 'bg-muted-transparent',
-        },
         size: {
-          sm: 'size-40 p-4',
-          md: 'size-48 p-4',
-          lg: 'size-56 p-4',
-          xl: 'size-72 p-4',
```
gamegee, line 20: [mid]: You seems to have dropped `padding: 4px` while not having dropped it for RNative

libs/ui-react/src/lib/Components/AvatarButton/AvatarButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <button
+      ref={ref}
+      type='button'
+      className={cn(
+        'group relative cursor-pointer rounded-full focus-visible:outline-2 focus-visible:outline-offset-4 focus-visible:outline-focus',
+        className,
+      )}
+      {...buttonProps}
+    >
```
gamegee, line 36: [low]: maybe to discuss with @LaurineDevergne , the AvatarButton will accept `disabled` state but has no styles associated.

**Probably to do now:**
We should probably at-least change the `cursor` property.

**To do later:**
Decide with Laurine if we need to add some opacity styles or other things..

FYI @aammami-ledger

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -1,40 +1,52 @@
...
+      const sizeMap: Record<Size, { size: number; padding: number }> = {
+        xs: { size: t.sizes.s24, padding: t.spacings.s4 },
         sm: { size: t.sizes.s40, padding: t.spacings.s4 },
         md: { size: t.sizes.s48, padding: t.spacings.s4 },
         lg: { size: t.sizes.s56, padding: t.spacings.s4 },
         xl: { size: t.sizes.s72, padding: t.spacings.s4 },
+        '2xl': { size: t.sizes.s128, padding: t.spacings.s4 },
+      };
+
+      const fallbackTextTypography = {
```
gamegee: [low]: Can add types like for other maps

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -1,40 +1,52 @@
...
+  fallbackColor?: string;
+  shouldFallback: boolean;
 }) => {
   return useStyleSheet(
     (t) => {
-      const backgroundColors: Record<Appearance, string> = {
-        gray: t.colors.bg.muted,
-        transparent: t.colors.bg.mutedTransparent,
-      };
+      const resolvedFallbackColor = fallbackColor;
```
gamegee: [low]: do you need that constant alias ? or could we directly use `fallbackColor`

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -78,23 +111,36 @@ export const Avatar = ({
...
+    <Text style={styles.fallbackText} selectable={false}>
+      {fallbackText}
+    </Text>
+  ) : (
+    <User
+      size={fallbackIconSizes[size]}
+      accessible={false}
+      testID='avatar-fallback-icon'
+    />
+  );
```
gamegee: [mid]: You could make it a sub-component in the same file that is not exposed.
It follows better composition best-practices of React, + React can better handle memoization better.

This is a pattern we start to do in components that grow in complexity (e.g. Input, LineChart, Scrubber, ...)

libs/ui-react/src/lib/Components/AvatarButton/AvatarButton.tsx
aquelemiguel: maybe `disabled:cursor-not-allowed` works for now?

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: good one. definitely not, it's a result of a refactor, pushed the change

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: while i think in this case we lose a bit of cleanliness because we need to forward quite a few props to that sub-component i can understand the idea of making it a pattern 👍 pushed it

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: great catch, previously this is what gave that border to the avatar
now that the border is inset, absolutely positioned on top of the image, we can remove in both libs

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
gamegee: Yes the only cons is props forwarding, and so re-writing/extending the props types

libs/ui-react/src/lib/Components/AvatarButton/AvatarButton.tsx
gamegee: Can start with it 👍

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
gamegee: alright, perfect

libs/ui-react/src/lib/Components/Avatar/resolveAvatarColor.ts
gamegee: You have a identity function exposed from design-core that is:
`cssVar()` 

It adds just typesafety on top of our css-tokens, and return the css var as output.

We are using it a lot in visualization, as we can't use tailwind on some svg-elements

libs/ui-react/src/lib/Components/AvatarButton/AvatarButton.tsx
aquelemiguel: reached out to laurine regardless, will reopen this convo if she updates the styles

libs/ui-react/src/lib/Components/Avatar/resolveAvatarColor.ts
aquelemiguel: i was unaware of this helper, this is a great one, pushed

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -43,28 +54,70 @@ const useStyles = ({
...
+          bottom: 0,
+        },
+        insetBorder: {
+          position: 'absolute',
+          top: 0,
+          left: 0,
+          right: 0,
+          bottom: 0,
           borderRadius: 9999,
+          borderWidth: 1,
```
aammami-ledger: This should be customisable (it's one of the requirements in the ticket)

aammami-ledger (changes requested):
I thinks you are missing one requirement which is:
- Border should be `default=1px` and customisable to X pixel

libs/ui-react/src/lib/Components/Avatar/types.ts
```diff
@@ -37,4 +30,24 @@ export type AvatarProps = {
...
+   * @optional
+   */
+  fallbackText?: string;
+  /**
+   * Ring appearance for the avatar.
+   * - `thin`: 1px ring using the icon border color
+   * - `thick`: 2px ring using the muted subtle transparent border color
+   * @optional
+   */
+  ringAppearance?: 'thin' | 'thick';
```
aammami-ledger:
```suggestion
  appearance?: 'thin' | 'thick';
```

I'd still go with `appearance` we don't have the `somethingAppearance` convention anywhere else in the codebase, and I can see it getting messy fast (e.g. `imgAppearance`, etc.). Not a fan of that direction.

libs/ui-rnative/src/lib/Components/Avatar/types.ts
```diff
@@ -12,15 +12,29 @@ export type AvatarProps = {
...
   alt?: string;
   /**
-   * The visual appearance of the avatar background: `gray` or `transparent`.
+   * The size variant of the avatar.
    * @optional
-   * @default transparent
+   * @default md
    */
-  appearance?: 'gray' | 'transparent';
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl';
```
aammami-ledger: should we add  the new sizes mapping in `getDotIndicatorProps` ?

libs/ui-rnative/src/lib/Components/AvatarButton/AvatarButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * <AvatarButton src="https://example.com/photo.jpg" alt="Open user menu" onPress={openMenu} />
+ */
+export const AvatarButton = ({
+  src,
+  alt,
+  size,
+  fallbackText,
+  fallbackColor,
+  disabled: disabledProp = false,
```
aammami-ledger, line 25: maybe add the shared disabled context ?

libs/ui-rnative/src/lib/Components/Avatar/types.ts
gamegee: Indeed, we might do the full mapping so consumers doesn't have issues with abstraction on top of avatar.

We had the discussion with @LucasWerey where they need to manually erase size type because `getDotIndicatorProps` is not fully mapped

libs/ui-rnative/src/lib/Components/Avatar/resolveAvatarColor.ts
```diff
@@ -0,0 +1,18 @@
+import { primitiveColorTokens } from '@ledgerhq/lumen-design-core';
+import {
+  resolveAvatarColorKey,
+  type AvatarColorKey,
+} from '@ledgerhq/lumen-utils-shared';
+
+const AVATAR_COLOR_TOKENS: Record<AvatarColorKey, string> =
+  primitiveColorTokens.light.avatar;
```
aammami-ledger, line 9: only `light` theme ? 👀

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -78,23 +139,30 @@ export const Avatar = ({
...
   const shouldFallback = !src || error;
-  const styles = useStyles({ appearance, size });
+  const styles = useStyles({
+    size,
+    fallbackColor,
+    shouldFallback,
+    appearance,
+  });
 
   const resolvedAlt = alt || t('components.avatar.defaultAriaLabel');
```
aammami-ledger, line 159: `alt` defaulted to Avatar 🤔 the alt || t('components.avatar.defaultAriaLabel') fallback never triggers

libs/ui-rnative/src/lib/Components/AvatarButton/types.ts
```diff
@@ -0,0 +1,15 @@
+import type { StyledPressableProps } from '../../../styles';
+import type { AvatarProps } from '../Avatar/types';
+
+export type AvatarButtonProps = Pick<
+  AvatarProps,
+  'src' | 'alt' | 'size' | 'fallbackText' | 'fallbackColor'
```
aammami-ledger: No appearance for the AvatarButton ?

libs/ui-react/src/lib/Components/Avatar/Avatar.tsx
```diff
@@ -6,35 +7,70 @@ import type { AvatarProps } from './types';
...
+  return fallbackText ? (
+    <span className={cn(avatarVariants.text({ size }), colorClass)}>
+      {fallbackText}
+    </span>
+  ) : (
+    <User
+      className={colorClass}
+      size={fallbackSizes[size]}
+      aria-label='Fallback Icon'
+      aria-hidden='true'
```
aammami-ledger: the label is dead since aria-hidden removes it from the a11y tree.
```suggestion
      aria-hidden='true'
```

libs/design-core/src/lib/themes/js/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from './enterprise/theme';
 export * from './ledger-live/theme';
 export * from './websites/theme';
+export * from './primitives/primitive.colors';
```
aammami-ledger, line 4: You need a release plan for `design-core`

libs/ui-rnative/src/lib/Components/Avatar/resolveAvatarColor.ts
aquelemiguel: either `light` or `dark` works, they're the same. i would've preferred just one instance as well to make the usage like this cleaner but for crypto color tokens we duplicated as well, so kept it for the time being

libs/ui-rnative/src/lib/Components/Avatar/Avatar.tsx
aquelemiguel: nice catch :)

libs/ui-rnative/src/lib/Components/AvatarButton/types.ts
aquelemiguel: forgot to spread the new prop indeed 😬
this PR got so gigantic it's very hard to track all moving parts...

libs/design-core/src/lib/themes/js/index.ts
aquelemiguel: added!

.nx/version-plans/version-plan-1783958823785-ui-rnative.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ledgerhq/lumen-ui-rnative': patch
+---
+
+feat(Avatar): add letter and background fallbacks + xs and 2xl sizes
```
aammami-ledger: We need to add the breaking change note: that visually the Avatar has changed and it has no stroke by default any more + breaking change for appearance.

I think we need to provide the migration plan for AI. Maybe Laurine can help do the mapping, something like: 

```
old appearance 'gray' is now appearance 'undefined'
old appearance 'transparent' is appearance 'thin' 
```


Also add that we have introduced a new AvatarButton.

libs/ui-rnative/src/lib/Components/Avatar/resolveAvatarColor.ts
aammami-ledger: Fair,  maybe worth adding comment noting the light/dark-identical assumption, so it doesn't silently break if palettes diverge later.

aammami-ledger (approved):

PR #803, opened 2026-07-17
DLS-917/split-table-into-compound

gamegee:


libs/ui-react/src/lib/Components/Table/types.ts
```diff
@@ -101,30 +101,69 @@ export type TableCellProps = {
...
+} & ComponentPropsWithRef<'div'>;
+
+export type TableCellContentDescriptionProps = {
+  /**
+   * The description content
+   */
+  children: ReactNode;
+  /**
+   * Custom classname
+   */
```
aquelemiguel, line 152: bit redundant, but if anything add @optional decorator i suppose

libs/ui-react/src/lib/Components/Table/Table.tsx
```diff
@@ -338,7 +342,11 @@ export const TableCell = ({
   );
 };
 
-const cellContentVariants = cva('flex min-w-0 items-center gap-12', {
+const [TableCellAlignProvider, useTableCellAlignContext] = createSafeContext<{
+  align: 'start' | 'end';
```
aquelemiguel: we could type this as `TableCellItemProps["align"]` to have one source of truth

aquelemiguel (approved):

PR #805, opened 2026-07-20
feat(DonutChart): Add interactivity

aammami-ledger:
image


libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -24,7 +69,14 @@ export function DonutChart({
...
         geometry={geometry}
         accessibilityLabel={accessibilityLabel}
+        activeId={activeId}
       />
+      <GestureDetector gesture={tap}>
+        <Animated.View
+          testID='donut-gesture-overlay'
+          style={StyleSheet.absoluteFill}
+        />
+      </GestureDetector>
```
aammami-ledger, line 76: using `GestureDetector` since 	`onPress` on Android has two separate bugs that both hit our donut slices:

onPress on a <G> (SVG group) doesn't reliably reach the tapped child — [react-native-svg#1321](https://github.com/react-native-community/react-native-svg/issues/1321)
onPress on a shape animated via Reanimated's useAnimatedProps often doesn't fire at all — [reanimated#2995](https://github.com/software-mansion/react-native-reanimated/issues/2995)

libs/ui-rnative-visualization/src/lib/Components/DonutChart/constants.ts
```diff
@@ -1,28 +1,66 @@
...
+    hitSlopRadius: 8,
   },
 } as const satisfies Record<string, DonutGeometry>;
 
+export const DONUT_INTERACTION = {
+  dimOpacity: 0.3,
+  opacityDurationMs: 150,
+  popDurationMs: 180,
+  popEasing: Easing.bezier(0.2, 0.8, 0.2, 1),
+} as const;
```
gamegee, line 39: [mid]: could we use design-core tokens ? Or is it purely custom ?
If so, it should probably be exposed as a hook that connect to the theme

libs/ui-react-visualization/src/lib/Components/DonutChart/utils.ts
```diff
@@ -10,6 +10,12 @@ export const DONUT_GEOMETRY: Readonly<Record<DonutSize, DonutGeometry>> =
 export const resolveSegmentColor = (segment: DonutSegment): string =>
   segment.color ?? chartConfig.donut.defaultSegmentColor;
 
+export const getDonutViewBox = (geometry: DonutGeometry): string => {
+  const { box, hoverOffset } = geometry;
+  const paddedBox = box + 2 * hoverOffset;
+  return `-${hoverOffset} -${hoverOffset} ${paddedBox} ${paddedBox}`;
+};
```
gamegee, line 17: [mid]: _FROM Claude_

 **Padded `viewBox` shrinks the ring at rest.** `getDonutViewBox` grows the viewBox by `2 * hoverOffset` while the `<svg>` `width`/`height` stay at `box`, so all content is scaled down by `box / (box + 2*hoverOffset)` — ~0.961 for md (168→174.72) and ~0.952 for sm. That means the donut renders visibly smaller *at rest*, not just when a slice pops out, so the rendered diameter no longer matches the Figma box sizes the `DONUT_GEOMETRY` comment claims to match.

Same applies to the RN `getDonutViewBox` in `constants.ts`.

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -8,21 +9,55 @@ export function DonutChart({
...
+      !(relatedTarget instanceof Node) ||
+      !event.currentTarget.contains(relatedTarget);
+    if (focusLeftRing) {
+      resetActiveId();
+    }
+  };
+
   return (
     <div
+      role='presentation'
```
gamegee, line 44: [low]: `role="presentation"` on a plain `<div>` is a no-op (a `div` has no implicit ARIA role), so it can read as if it's doing something. The meaningful additions here are `onMouseLeave`/`onBlur`. Safe to drop unless it's satisfying a specific linter issue.

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -1,20 +1,62 @@
...
+  });
+
   const arcs = useMemo(() => buildArcs(series, geometry), [series, geometry]);
 
+  const handleSegmentPress = useCallback(
+    (id: string) => {
+      setActiveId(activeId === id ? null : id);
+    },
+    [activeId, setActiveId],
+  );
```
gamegee, line 36: [mid]: Minor perf: `handleSegmentPress` closes over `activeId`, so it (and the downstream `handleTap` + memoized `tap` gesture) is recreated on every selection, forcing `GestureDetector` to re-attach. `useControllableState`'s setter supports functional updates, so this is avoidable:

```ts
const handleSegmentPress = useCallback(
  (id: string) => {
    setActiveId((prev) => (prev === id ? null : id));
  },
  [setActiveId],
);
```

That keeps the gesture stable across selections.

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -24,7 +66,14 @@ export function DonutChart({
...
         geometry={geometry}
         accessibilityLabel={accessibilityLabel}
+        activeId={activeId}
       />
+      <GestureDetector gesture={tap}>
+        <Animated.View
+          testID='donut-gesture-overlay'
+          style={StyleSheet.absoluteFill}
+        />
+      </GestureDetector>
```
gamegee, line 76: [low]: The gesture overlay is a static `Animated.View` with `StyleSheet.absoluteFill` and no animated props. A plain `View` works with `GestureDetector` here - no need for `Animated.View` unless there's a reason to keep it.

libs/ui-rnative-visualization/src/lib/Components/DonutChart/types.ts
```diff
@@ -27,4 +27,10 @@ export type DonutChartProps = {
    * @default 'Donut chart'
    */
   accessibilityLabel?: string;
+  /** Controlled active segment id. Active is data-driven, not a state variant. */
+  activeId?: string | null;
+  /** Default active segment id for uncontrolled mode. */
+  defaultActiveId?: string | null;
+  /** Fired when the active segment changes (tap or external reset). */
```
gamegee, line 34: [low]: Doc nitpick: "(tap or external reset)" - on RN there's no reset-on-leave; `activeId` only changes by tapping a new slice or re-tapping the active one (web resets on `mouseleave`/blur). Consider rewording to match the tap-toggle semantics, e.g. "(tap to select, tap again to deselect)", consistent with the sandbox `Section` title.

libs/ui-react-visualization/src/lib/config.ts
```diff
@@ -121,20 +121,27 @@ export const chartConfig = {
   donut: {
     defaultSegmentColor: cssVar('var(--background-muted-strong)'),
     emptyRingColor: cssVar('var(--background-muted)'),
+    hover: {
+      dimOpacity: 0.3,
+      opacityTransition: 'opacity 150ms ease',
+      popTransition: 'transform 180ms cubic-bezier(0.2, 0.8, 0.2, 1)',
```
gamegee, line 127: [low]: just noticed we don't really have linting here, but that's fine

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
aammami-ledger: image
yes it is to avoid a linter issue

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutRing.tsx
```diff
@@ -1,16 +1,56 @@
...
+      onMouseEnter={handleSegmentEnter}
+      onFocus={handleSegmentEnter}
+      style={{
+        cursor: 'pointer',
+        opacity,
+        transform: `translate(${x}px, ${y}px)`,
+        transition: segment.hoverEnabled
+          ? `${hover.opacityTransition}, ${hover.popTransition}`
+          : undefined,
+      }}
```
gamegee, line 50: [low] can't use CVA to compute that ?

nvm, maybe we did the same for LineChart, so we could keep it - its just it seems most properties here could be just tailwind design-tokens

libs/ui-rnative-visualization/src/lib/Components/DonutChart/constants.ts
aammami-ledger: It's purely custom, I also followed what has been done on line-chart animations, all was inlined none is imported from design-core

gamegee (approved):

libs/ui-react-visualization/src/lib/Components/DonutChart/utils.ts
aammami-ledger: it's intended, that's how it's done on figma

PR #806, opened 2026-07-20
feat(MediaBanner): support click/hover behaviour

aquelemiguel:
Closes [DLS-834](https://ledgerhq.atlassian.net/browse/DLS-834).

[DLS-834]: https://ledgerhq.atlassian.net/browse/DLS-834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-react/src/lib/Components/MediaBanner/MediaBanner.tsx
```diff
@@ -21,57 +70,70 @@ export const MediaBanner = ({
...
-      </div>
-      <div className='relative w-128'>
-        {showImage && (
-          <img
-            src={imageUrl}
-            alt=''
-            aria-hidden
-            className='absolute inset-0 size-full object-cover'
-            onError={() => setImageLoadError(true)}
+      <DisabledProvider value={{ disabled }}>
```
aammami-ledger: Is it expected that disabled banner also disables its close button ? maybe the close button should be outside the `DisabledProvider` ?

libs/ui-react/src/lib/Components/MediaBanner/MediaBanner.tsx
```diff
@@ -21,57 +70,70 @@ export const MediaBanner = ({
...
+
   useEffect(() => {
     setImageLoadError(false);
   }, [imageUrl]);
 
   const showImage = imageUrl && !imageLoadError;
 
   return (
-    <div
+    <button
```
aammami-ledger: `<button>` containing the close InteractiveIcon (`<button>`) → nested buttons = invalid DOM + React hydration error: maybe when the Card is disabled let's drop the close button ?

aammami-ledger: Also it should be a button only when is has `onclick` no ? otherwise accessibility wise it will be announced as a button and tab-focusable while doing nothing

libs/ui-rnative/src/lib/Components/MediaBanner/MediaBanner.tsx
```diff
@@ -61,51 +73,66 @@ export function MediaBanner({
...
+      style={({ pressed }) => [
+        styles.container,
+        pressed &&
+          !disabled &&
+          onPress && { backgroundColor: t.colors.bg.surfacePressed },
+        style,
+      ]}
+      onPress={onPress}
+      disabled={disabled}
+      accessibilityState={{ disabled }}
```
aammami-ledger: missing `accessibilityRole='button'`

libs/ui-react/src/lib/Components/MediaBanner/MediaBanner.tsx
```diff
@@ -21,57 +70,70 @@ export const MediaBanner = ({
...
       ref={ref}
+      type='button'
+      disabled={disabled}
+      onClick={disabled ? undefined : onClick}
```
aammami-ledger: the disabled prop already prevents the handler from firing
```suggestion
      onClick={onClick}
```

aammami-ledger (commented):
I think the img should also has reduced opacity when banner is disabled, WDYT ? 

Image

libs/ui-react/src/lib/Components/MediaBanner/MediaBanner.tsx
```diff
@@ -71,7 +134,7 @@ export const MediaBanner = ({
           aria-label={closeAriaLabel || t('components.banner.closeAriaLabel')}
         />
```
aammami-ledger, line 91: Can we move the close button to outside the wrapper ? to avoid having a button inside a button when interactive ?

libs/ui-react/src/lib/Components/MediaBanner/MediaBanner.tsx
```diff
@@ -21,43 +70,57 @@ export const MediaBanner = ({
...
-            src={imageUrl}
-            alt=''
-            aria-hidden
-            className='absolute inset-0 size-full object-cover'
-            onError={() => setImageLoadError(true)}
-          />
-        )}
-        <div className='absolute inset-0 bg-linear-[45deg] from-black/0 from-67% to-black/80' />
-      </div>
+      <DisabledProvider value={{ disabled }}>
```
aammami-ledger: honestly I am not even sure it was the right suggestion from my side UX wise it feels weird to be able to close the banner when disabled when testing it, maybe we should also visibly disable the CloseButton when disabled ? 

I am at the stage where I am even questioning why a banner would be disabled ? 🤔

libs/ui-react/src/lib/Components/MediaBanner/MediaBanner.tsx
```diff
@@ -32,42 +47,49 @@ export const MediaBanner = ({
 
   const showImage = imageUrl && !imageLoadError;
 
+  const Wrapper = (onClick ? 'button' : 'div') as 'button';
```
aammami-ledger: Can we avoid this cast please ?  it will let consumers pass button props to a div when no onClick
```suggestion
const Wrapper = (onClick ? 'button' : 'div');
```

aammami-ledger (approved):

libs/ui-react/src/lib/Components/MediaBanner/MediaBanner.tsx
aammami-ledger: Seeing how we handle this on other components is via asChild/slot (that we want to deprecate at some point) 
So not sure how to properly fix this 🤔 

cc @gamegee

aquelemiguel: it's a type-only nuance, because to potentially assign `onClick` and `ref` the type of wrapper needs to be `HTMLButtonElement`. only alternative would be:

* have MediaBannerProps types be a discriminated union of `ComponentsWithRef<'button'>` and `ComponentsWithRef<'div'>`, don't leak `onClick` onto div anymore
* do button/div branching instead of the Wrapper component

safer, but harder to reason IMO when looking at the code

gamegee: Could be the same pattern as the `card` by using `getButtonA11yProps`, wdyt ?

gamegee: So yes indeed, having only one element is easier regarding typesafety.
And web is so flexible that I would suggest to do so.

---
That being said react-native is less flexible, and in few components we are rendering twice, either a `Box` or a `Pressable` doing some conditional rendering. 

I am thinking that at some point, it would be nice to have an internal utility component that handle that, to prevent code duplication and conditionals

---
Dumb example of API

### NEW
```tsx
<BoxOrPressable
  isPressabled={Boolean(onClick)}
  render={({ pressed }) => { // pressed will just be false all time when its a Box
    <InnerContent
      propA='A' 
      propB='B' 
     />
  }}
/>
```

### CURRENT
```tsx
  if (onClick) {
    return (
      <Pressable>
        {({ pressed }) => 
          <InnerContent
           propA='A' 
           propB='B' 
      />}
      </Pressable>
    );
  }

  return (
    <Box>
      <InnerContent 
        propA='A' 
        propB='B' 
      />
    </Box>
  );
```

any thoughts ?

PR #810, opened 2026-07-22
feat(BottomSheet): support BottomSheetFooter

aquelemiguel:
Closes [DLS-826](https://ledgerhq.atlassian.net/browse/DLS-826).

## Considerations

* Gorhom's BottomSheetView is `position: absolute; top: 0` meaning it cannot coexist as a sibling of the header as it overlaps it. Our custom BottomSheetView extends Gorhom's.

```tsx
<BottomSheet>
  <BottomSheetHeader />
  <BottomSheetView> { /** <-- this container is placed right on top of the header! */ }
     <Text>Hello</Text>
  </BottomSheetView>
</BottomSheet>
```

* For dynamic sizing to work, a Gorhom-aware wrapper is required. Everything must be inside a Gorhom wrapper otherwise it won't be taken into account when calculating the dynamic height.

```tsx
<BottomSheet enableDynamicSizing snapPoints={null}>
  <BottomSheetView> { /** <-- this is mandatory, either our custom wrapper or directly Gorhom's  */ }
    <Header /> { /** <-- must be a child of view  */ }
    <Text>Hello</Text>
    <Footer /> { /** <-- must be a child of view  */ }
  </BottomSheetView>
</BottomSheet>
```

* Still for dynamic sizing, it's not possible to add multiple `BottomSheetView`s like: 

```tsx
<BottomSheet enableDynamicSizing snapPoints={null}>
  <BottomSheetView>
    <Header />
    <BottomSheetView> { /** <-- not possible, once again will be absolutely positioned to the top  */ }
      <Text>Hello</Text>
    </BottomSheetView>
  </BottomSheetView>
</BottomSheet>
```

* Therefore we are missing a content wrapper that does not extend Gorhom's BottomSheetView but instead is a styled custom Box container. Thus BottomSheetContent was added:

```tsx
<BottomSheet enableDynamicSizing snapPoints={null}>
  <BottomSheetView> 
    <Header />
   { /** BottomSheetContent already has consistent paddings in place  */ }
    <BottomSheetContent lx={{ flexDirection: 'column', gap: 's12' }}>
       <Text>Hello</Hello>
    </BottomSheetContent>
    <Footer>
  </BottomSheetView>
</BottomSheet>
```

* Note that our custom BottomSheetView already has horizontal and vertical padding, so to prevent double horizontal padding there's internal wiring via contexts to understand that if BottomSheetContent is a child of BottomSheetView, its `paddingHorizontal` should be dropped.

* Also, our new Footer already includes `paddingBottom`, so again, internal wiring so that if there's a footer inside the bottom sheet, BottomSheetContent drops its bottom padding.

* This way we can have all combinations of bottom sheet setup: with/without dynamic sizing + with/without footer while maintaining spacing consistency. 👍

## Demo

https://github.com/user-attachments/assets/eb36e2ea-e474-4406-8c89-c74a6e909dc4

[DLS-826]: https://ledgerhq.atlassian.net/browse/DLS-826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/app-sandbox-rnative/src/app/(blocks)/BottomSheets.tsx
```diff
@@ -186,9 +308,30 @@ export default function BottomSheets() {
...
+        onPress={() => footerFlatListRef.current?.present()}
+      >
+        Open
+      </Button>
+      <Button
+        appearance='base'
+        onPress={() => footerDynamicRef.current?.present()}
+      >
+        Open
+      </Button>
```
gamegee, line 328: [mid]: we could add informative label on each button, so we know which usecase its testing

libs/ui-rnative/src/lib/Components/BottomSheet/BottomSheet.stories.tsx
```diff
@@ -146,7 +148,7 @@ export const Base: Story = {
                   voluptates eum, sunt asperiores quod.
                 </Text>
               ))}
-            </Box>
+            </BottomSheetContent>
```
gamegee, line 196: [low]: we already discussed it in the past with @aammami-ledger - would be nice to have a story with bottom-sheet already open, so the ui-regression could catch any spacing issue if exist.

I believe it could be one good addition, wdyt ?

libs/ui-rnative/src/lib/Components/BottomSheet/Scrollables.tsx
```diff
@@ -15,16 +16,28 @@ import type {
...
         flex: 1,
         paddingHorizontal: t.spacings.s16,
-        paddingBottom: t.spacings.s16,
+        paddingBottom: hasFooter ? 0 : t.spacings.s16,
       },
     }),
-    [],
+    [hasFooter],
   );
 };
```
gamegee: [mid]: why not directly calling the `useHasFooter` here ? instead of duplicating

gamegee (approved):

PR #811, opened 2026-07-24
[D] feat(DonutChart): add `renderCenter` and `renderCenterActive`

aammami-ledger:
https://github.com/user-attachments/assets/155d458c-5273-47f1-adcc-dae3c44ca009


libs/ui-react-visualization/src/lib/Components/DonutChart/types.ts
```diff
@@ -55,4 +57,31 @@ export type DonutChartProps = {
...
+    activeSegment: DonutSegment & { percent: number };
+  }) => ReactNode;
+};
+
+export type DonutChartTitleProps = ComponentPropsWithRef<'div'> & {
+  /**
+   * The title's own typography scale.
+   * @default 'md'
+   */
+  size?: DonutSize;
```
aquelemiguel: should this be typed as DonutSize or rather as DonutTitleSize (even if the type union is the same), because then in

```tsx
const TITLE_TYPOGRAPHY: Record<DonutSize, Record<DonutSize, string>> = {
  md: { md: 'heading-1-semi-bold', sm: 'heading-2-semi-bold' },
  sm: { md: 'heading-4-semi-bold', sm: 'body-2-semi-bold' },
};
```

it'd be a bit more clear what these combinations refer to: `Record<DonutSize, Record<DonutTitleSize, string>>`

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
```diff
@@ -0,0 +1,40 @@
...
+  });
+  const maxWidth = getCenterMaxWidth(DONUT_GEOMETRY[donutSize]);
+
+  return (
+    <div
+      ref={ref}
+      style={{ maxWidth, ...style }}
+      className={cn(
+        'flex items-center gap-2 overflow-hidden text-muted',
+        '[&>.truncate]:min-w-0 [&>.truncate]:flex-1',
```
aquelemiguel: so AFAIK this forces the direct child of this to use the truncate class otherwise truncation won't apply, but what if the direct child is a flex container, for instance, to layout an icon and text?

if this is expected to truncate by default, could we not ditch that `[&>.truncate]` selector?

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
```diff
@@ -0,0 +1,40 @@
+import { cn } from '@ledgerhq/lumen-utils-shared';
+
+import { useDonutSizeContext } from './donutSizeContext';
+import type { DonutChartDescriptionProps, DonutSize } from './types';
+import { DONUT_GEOMETRY, getCenterMaxWidth } from './utils';
+
+const DESCRIPTION_TYPOGRAPHY: Record<DonutSize, string> = {
+  md: 'body-3',
+  sm: 'body-4',
+};
```
aquelemiguel: a bit of a type nitpick is that i believe this is a better definition:

```tsx
const DESCRIPTION_TYPOGRAPHY = {
  md: 'body-3',
  sm: 'body-4',
} as const satisfies Record<DonutSize, string>;
```

1. makes the object immutable
2. you get validation via the satisfies keyword
3. does not broaden the type to be a generic string:

```tsx
const res = DESCRIPTION_TYPOGRAPHY[donutSize];

res: string // currently
res: "body-3" | "body-4" // with const + satisfies
```

(same idea in TITLE_TYPOGRAPHY)

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -58,6 +74,32 @@ export function DonutChart({
...
       />
+      {hasCenter && (
+        <div
+          data-testid='donut-center'
+          className='pointer-events-none absolute inset-0 flex items-center justify-center'
+        >
+          <DonutSizeProvider value={{ size }}>
+            {useAnimatedCenter ? (
+              <DonutChartAnimatedCenter
+                activeSegment={activeSegment}
```
aquelemiguel, line 84: i had noticed this on the previous PR but didn't get the chance to point out: i believe the "hitslop / hitbox" for hovering segments could be expanded a bit so cycling through the segments by drawing circles with the mouse is a bit easier, i find it slightly difficult to alternate between them

slightly exaggerated demo:

https://github.com/user-attachments/assets/838120c4-f281-499a-98c5-de65ba90bc4a

libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.mdx
```diff
@@ -85,7 +85,36 @@ slot can react to the same active segment.
...
 <Canvas of={DonutChartStories.Interactive} />
 
-<Canvas of={DonutChartStories.Controlled} />
+
+## Center content
+
+`renderCenter` renders content in the middle of the ring from `DonutChartTitle`
+and `DonutChartDescription`. Layout and pointer events are handled internally.
+Used alone, its `activeSegment` param (enriched with `percent`, `null` when
+nothing is active).
```
aquelemiguel: is something missing in this sentence? used alone, its `activeSegment` param...?

libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.mdx
```diff
@@ -85,7 +85,36 @@ slot can react to the same active segment.
...
+
+Pass `renderCenterActive` alongside it to crossfade between resting and active
+layouts on hover instead: `renderCenter` then always gets `activeSegment: null`
+(resting only), while `renderCenterActive` handles the active layout.
+
+Both `DonutChartTitle` and `DonutChartDescription` cap their width to the
+ring's inner diameter, so long content truncates instead of overlapping the
+ring. In `DonutChartDescription`, mark the truncatable part of `children`
+(e.g. the label, not a trailing icon) with a `truncate` class so it shrinks
+while the icon stays fixed-size.
```
aquelemiguel: i think this is a bit too implementation heavy, it could state the purpose of the center content better, potential usage as well

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
aammami-ledger: I've dropped it already. It was unnecessary

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartAnimatedCenter.tsx
```diff
@@ -0,0 +1,76 @@
...
+    transition: centerTextTransition,
+    opacity: isActive ? 0 : 1,
+    transform: `translateY(${isActive ? -SLIDE_PX : 0}px)`,
+  };
+
+  const activeStyle = {
+    transition: centerTextTransition,
+    opacity: isActive ? 1 : 0,
+    transform: `translateY(${isActive ? 0 : SLIDE_PX}px)`,
+  };
```
gamegee: [low]: use `useMemo`

libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.stories.tsx
```diff
@@ -156,3 +161,112 @@ export const Controlled: Story = {
...
+          <DonutChartDescription>
+            <span className='truncate'>{activeSegment.label}</span>
+            <InteractiveIcon
+              iconType='stroked'
+              icon={ChevronRight}
+              size={16}
+              aria-label={`View ${activeSegment.label} details`}
+            />
+          </DonutChartDescription>
+        </>
```
gamegee: [low]: I guess the `renderCenter` & `renderCenterActive` functions will apply defaults padding & gap & alignments.
How consumers could erase these styles ?

I usually prefer that we don't render a fragment as child, but a top-level component, so consumers could override styles.

What about having:

```tsx
renderCenter={() => 
  <DonutChartCenter>
    <DonutChartTitle>
    <DonutChartDescription>
  </DonutChartCenter>
}
```

A bit more verbose, but also more explicit

gamegee: I put it as low because users can also do that:

```tsx
renderCenter={() => 
  <div>
    <DonutChartTitle>
    <DonutChartDescription>
  </div>
}
```

But its mostly the fragment render that is IMO not usual

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <div
+      ref={ref}
+      style={{ maxWidth, ...style }}
+      className={cn(
+        'flex items-center gap-2 truncate overflow-hidden text-muted',
+        donutSize === 'sm' && '-mt-4',
+        DESCRIPTION_TYPOGRAPHY[donutSize],
+        className,
+      )}
```
gamegee: [low]: can use CVA when you start to have multiple conditions

aquelemiguel (approved):

PR #815, opened 2026-07-28
[M] feat(DonutChart): add renderCenter and renderCenterActive

aammami-ledger:
Screenshot 2026-07-29 at 10 21 13


gamegee (commented):
[low/mid]: we are missing unit tests on sub-components.
I would suggest either: 
- add test for each of them
- add higher-level tests in `DonutChart.test.tsx`

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChartAnimatedCenter.tsx
```diff
@@ -0,0 +1,98 @@
...
+  return (
+    <View style={[styles.container, { width: contentWidth }]}>
+      <Animated.View
+        accessibilityElementsHidden={isActive}
+        importantForAccessibility={isActive ? 'no-hide-descendants' : 'auto'}
+        pointerEvents={isActive ? 'none' : 'auto'}
+        style={[
+          hasEverActivated ? styles.overlay : styles.primary,
+          restingStyle,
+        ]}
```
gamegee, line 41: [mid]: **Container is sized by only the active pane after first activation.**

On web, `DonutChartAnimatedCenter` uses a CSS grid with both panes in `grid-area: 1/1`, so the container always sizes to the max of resting/active content. Here, once activated the resting pane switches to `styles.overlay` (`absoluteFillObject`), leaving the active pane as the only in-flow child driving height - and the container has a fixed `width` but no matching height anchor. If resting content is taller than active content, layout/centering ends up driven by the active pane.

**Solution**
Make the resting pane always in-flow and the active pane always the overlay. Drop hasEverActivated entirely:
```tsx
return (
  <View style={[styles.container, { width: contentWidth }]}>
    <Animated.View
      accessibilityElementsHidden={isActive}
      importantForAccessibility={isActive ? 'no-hide-descendants' : 'auto'}
      pointerEvents={isActive ? 'none' : 'auto'}
      style={[styles.primary, restingStyle]}
    >
      {renderResting()}
    </Animated.View>
    {shown && (
      <Animated.View
        accessibilityElementsHidden={!isActive}
        importantForAccessibility={isActive ? 'auto' : 'no-hide-descendants'}
        pointerEvents={isActive ? 'auto' : 'none'}
        style={[styles.overlay, activeStyle]}
      >
        {renderActive(shown)}
      </Animated.View>
    )}
  </View>
);
```

gamegee (commented):
Might miss the version plan for react small changes

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChartAnimatedCenter.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const restingStyle = useAnimatedStyle(() => ({
+    opacity: 1 - progress.value,
+    transform: [{ translateY: -progress.value * transitionSlideDistance }],
+  }));
+
+  const activeStyle = useAnimatedStyle(() => ({
+    opacity: progress.value,
+    transform: [{ translateY: (1 - progress.value) * transitionSlideDistance }],
+  }));
```
gamegee: [low]: we might split into separate hooks, like we do sometimes (separation of concerns)

gamegee (approved):
✅ pre-approved

PR #817, opened 2026-07-30
[D][M] feat(DonutChart): implement loading state

aammami-ledger:
https://github.com/user-attachments/assets/356a8b57-d6b8-4ae4-92e6-8b0cb54dafcd


libs/ui-react-visualization/src/lib/Components/DonutChart/DonutRing.tsx
```diff
@@ -52,19 +53,55 @@ const RingSegment = ({
...
+            style={{
+              animation: animationStyle,
+              animationDelay: getSegmentDelay(segment.midAngle),
+            }}
+          />
+        ))}
+      </g>
+    </>
+  );
+};
```
gamegee, line 103: [low]: What about splitting this file into a sub-folder `DonutRingLoading`, this folder could also contain the hook, so its scoped under a same folder. Easy to reason about, self-encapsulated and testable

gamegee: Note: that being said LineChart is not splitting the files either, so up to you.
If you do for one, maybe do for both

libs/ui-react-visualization/src/lib/Components/DonutChart/utils.ts
```diff
@@ -110,12 +116,40 @@ const snapHalfCircle = <T>(datum: PieArcDatum<T>): PieArcDatum<T> => {
...
 
-/** Full, gapless ring path used for the empty ring. */
-export const buildEmptyRingPath = (geometry: DonutGeometry): string => {
-  const emptyRingArc = arc<unknown>()
+/**
+ * Fixed placeholder arcs shared by the empty ring and the loading wave: a
+ * full ring (values sum to 100) of unequal segments separated only by the
+ * same small `padAngle` gaps as real segments — no missing arc.
+ */
+export const buildPlaceholderArcs = (
```
gamegee: [mid]: I believed I've renamed recently `Arcs` to `Segments` everywhere, it makes more sense to have the same term everywhere

libs/ui-react-visualization/src/lib/config.ts
```diff
@@ -139,6 +141,15 @@ export const chartConfig = {
       popTransition: 'transform 180ms cubic-bezier(0.2, 0.8, 0.2, 1)',
       centerTextTransition: 'opacity 180ms ease, transform 180ms ease',
     },
+    placeholder: {
+      segmentValues: [25, 19, 15, 13, 10, 10, 8],
+    },
+    loading: {
+      durationInSeconds: 2,
```
gamegee, line 146: [low]: stick to `durationInMs` - we are using `ms` everywhere

libs/ui-rnative-visualization/src/lib/config.ts
```diff
@@ -125,6 +124,14 @@ export const chartConfig = {
       popDurationMs: 180,
       popEasing: Easing.bezier(0.2, 0.8, 0.2, 1),
     },
+    placeholder: {
+      segmentValues: [25, 19, 15, 13, 10, 10, 8],
+    },
+    loading: {
+      durationInSeconds: 2,
```
gamegee: same

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutRing.tsx
```diff
@@ -83,22 +92,82 @@ const EmptyRing = ({
 }: {
   geometry: DonutGeometry;
   color: string;
-}) => (
-  <Path testID='donut-empty' d={buildEmptyRingPath(geometry)} fill={color} />
-);
+}) => {
+  const arcs = useMemo(() => buildPlaceholderArcs(geometry), [geometry]);
```
gamegee: [mid] same, rename to segments ?

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutRing.tsx
aammami-ledger: hmm, i am leaving as-is. The three inner components are the branches of one switch that lives in the same file, so it reads as a single concept, they do not need any type/tests ect so not sure I see value in seperation

gamegee (approved):

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutRing.tsx
gamegee: ok

PR #818, opened 2026-07-31
chore: add prepare segments (group + sort)

gamegee:
Signed-off-by: Simon Bruneaud <simon.bruneaud@ledger.fr>

### DIFF
--- a/libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.mdx
+++ b/libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.mdx
@@ -130,9 +130,10 @@ ring:
 - **Sorting** — orders segments by `value` descending, so the largest slice
   starts at 12 o'clock. Sorting always applies, including when grouping is off.
 - **Grouping** — folds the long tail into a single "Other" slice, keeping the
-  top `maxSegments` and, optionally, anything at or above `minShare`. A tail of
-  a single segment is left alone, since an "Other" slice worth exactly that one
-  segment says less than the segment itself.
+  segments that are both within the top `maxSegments` and at or above `minShare`
+  (a fraction of the total, so `0.04` for 4%). A tail of a single segment is left
+  alone, since an "Other" slice worth exactly that one segment says less than the
+  segment itself.
 
 It returns `segments` (chart-ready, with `other` appended) and `others` (the raw
 segments that were folded in, handy for a "+N more" legend). Only `other.label`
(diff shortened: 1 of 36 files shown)


libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.mdx
```diff
@@ -103,19 +107,71 @@ Return a top-level `DonutChartCenter` wrapping `DonutChartTitle` and/or
...
+  minShare?: number;
+  /** Shape of the aggregate slice; its `value` is always computed. */
+  other: {
+    label: string;
+    /** @default 'other' */
+    id?: string;
+    /** Optional; defaults to the chart's neutral segment color. */
+    color?: string;
+  };
+};
```
aquelemiguel, line 160: maybe we'd move this code block to the implementation tab? given we don't have one right now, maybe discard for now

libs/ui-react-visualization/src/lib/Components/DonutChart/types.ts
```diff
@@ -32,18 +32,79 @@ export type DonutArc = {
...
+/** How `useDonutSeries` folds a raw series' long tail into a single slice. */
+export type DonutSeriesOptions = {
+  /**
+   * Segments below this share of the total are grouped into the "other" segment.
+   * A fraction of the total, not a percentage: pass `0.04` for 4%. `0` means no
+   * minimum. The top segment is always kept, so `1` or above groups everything
+   * but the largest rather than collapsing the chart.
+   * @default 0
+   */
+  minShare?: number;
```
aquelemiguel: this prop's name could be a bit more self-describing. something like `otherThreshold` could work.

libs/ui-react-visualization/src/lib/Components/DonutChart/types.ts
```diff
@@ -32,18 +32,79 @@ export type DonutArc = {
...
+    /** Optional color override; defaults to the chart's neutral segment color. */
+    color?: DonutSegment['color'];
+  };
+};
+
+export type DonutSeriesResult = {
+  /**
+   * Kept slices, sorted by value descending, plus the appended `other` slice.
+   */
+  segments: DonutSegment[];
```
aquelemiguel, line 86: slices vs segments on the docs, we should probably just use `segments` as to not confuse the consumer.

libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.mdx
gamegee: In visualization libs we took another approach -> Mixing both `implem` & `overview` tab

This is for the reason that, LLM have access to the HTML of storybook for now, and its easier to just have all the information in a one-pager.

I feel this is the direction we want to take, even if legacy stories have the tabs.

libs/ui-react-visualization/src/lib/Components/DonutChart/types.ts
gamegee: Agreed will fix

libs/ui-react-visualization/src/lib/Components/DonutChart/types.ts
gamegee: -> Or `groupThreshold` 🤔 
`minShare` was the suggested name from cursor & claude, but indeed it requires some JSDoc info to fully understand

aquelemiguel: i like `groupThreshold` also! it was "share" that didn't immediately register as percentage to me, but perhaps because english isn't my native lang

aquelemiguel (approved):

libs/ui-react-visualization/src/lib/Components/DonutChart/types.ts
gamegee: agreed

PR #820, opened 2026-08-04
feat(DonutChart): add reveal animation

aquelemiguel:
Closes [DLS-867](https://ledgerhq.atlassian.net/browse/DLS-867).

## Mobile demo

Slowed down to 5s animation to be easier to visualize the animation and easing here in the demo.

**Note that the animation is only available on iOS. Can confirm after testing that animation will not work on Android (tested on Pixel 9a simulator). Check Copilot's [comment](https://github.com/LedgerHQ/lumen/pull/820#discussion_r3721845540).** 

https://github.com/user-attachments/assets/62d6da94-2b42-46e2-86f9-5caeeae7b0ab

### Reduced motion

Respects the "reduce motion" accessibility setting by querying the system via Reanimated's `useReducedMotion` hook. If this setting is enabled, the reveal animation does not play.

https://github.com/user-attachments/assets/c086fb17-0fbd-4320-9024-8376f38cde8b

[DLS-867]: https://ledgerhq.atlassian.net/browse/DLS-867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-rnative-visualization/src/lib/config.ts
```diff
@@ -124,6 +124,9 @@ export const chartConfig = {
       popDurationMs: 180,
       popEasing: Easing.bezier(0.2, 0.8, 0.2, 1),
     },
+    reveal: {
+      durationMs: 1000,
```
gamegee, line 128: [mid]: It seems we are using custom durations, should we instead move that to the `resolveChartTokens` and consume the theme directly instead ?


@aammami-ledger fyi


That being said probably `180ms` doesn't exist

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutRing.tsx
```diff
@@ -81,29 +100,33 @@ export const DonutRing = ({
...
-              onSegmentEnter={onSegmentEnter}
-            />
-          ))
-        ) : (
-          <EmptyRing geometry={geometry} />
-        )}
-      </g>
-    </svg>
+    // eslint-disable-next-line better-tailwindcss/no-unknown-classes
+    <div className='donut-ring-reveal'>
```
gamegee: [mid]: I would suggest to split a bit more the components, for the LineChart we have a `RevealAnimationProvider` which encapsulate the logic.

Its more complicated in the LineChart, so its both a component + a provider.

But in your case having a `<RevealAnimation>` wrapper, on both web & mobile would makes sense. Especially in mobile, it would keep the DonutRing code simpler when we will add more features.

aammami-ledger (commented):
Not running on simulator: 
Image

aquelemiguel:
@aammami-ledger thank you for the report, i was missing the worklet decorator in the `computeRevealClipPath` function. fixed in https://github.com/LedgerHQ/lumen/pull/820/commits/c3c1b489783a41b23efa9a88ddb491eab1198284

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutRing.tsx
aquelemiguel: excellent idea, that wrapper. tyvm! added in https://github.com/LedgerHQ/lumen/pull/820/commits/efc05a73c8696cac7477ec8d450d940189658670

gamegee (commented):
Video of Reveal-anim of `1000ms` seems a bit too much

But maybe its just due to the Bezier curve (should be ease-out) ?

gamegee (approved):
pre-approved, clean LGTM ✅ 
Few minor coms left

aammami-ledger (changes requested):
Spotted small regression compared to main: the Arc is now cropped when you select it on mobile:

you branch:
Image

main:
Image

aquelemiguel:
@aammami-ledger amazingly caught, fixed in https://github.com/LedgerHQ/lumen/pull/820/commits/9108ed75081eee50ddc7aef04f47f8183a48c0c2, should be good again

image

aammami-ledger (approved):

gamegee (approved):

PR #823, opened 2026-08-06
feat(LineChart,DonutChart): create util for tweaking contrast unsafe colors

aquelemiguel:
Closes [DLS-870](https://ledgerhq.atlassian.net/browse/DLS-870).

### DonutChart

https://github.com/user-attachments/assets/90b951df-6173-49d0-af43-b6766f18df51

### LineChart

https://github.com/user-attachments/assets/1add5ee5-9044-4e76-9801-ec37932fb992

image

image

[DLS-870]: https://ledgerhq.atlassian.net/browse/DLS-870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/design-core/src/lib/colorContrast/colorContrast.ts
```diff
@@ -0,0 +1,114 @@
...
+ * darkened further.
+ *
+ * Both `color` and `bgColor` must be 6-digit hex strings (e.g. `#aabbcc`).
+ * CSS variables and `rgb()` values are not supported.
+ *
+ * @param color - The foreground hex color to adjust.
+ * @param bgColor - The background hex color to contrast against.
+ * @param threshold - Minimum lightness delta to enforce. Defaults to `20`.
+ */
+
```
gamegee: ✂️ space

libs/design-core/src/lib/colorContrast/colorContrast.ts
```diff
@@ -0,0 +1,114 @@
+type HslValue = {
+  h: number;
+  s: number;
+  l: number;
+};
+
+function hexToHsl(hex: string): HslValue {
```
gamegee: I would split each fn to its own fild, maybe at some point these are helpers we will expose publicly in the package

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -12,18 +16,69 @@ import {
...
+  const bgColor = useContrastBgColor();
+  return <DonutChartInner {...props} bgColor={bgColor} />;
+}
+
+export function DonutChart(props: Readonly<DonutChartProps>) {
+  if (props.ensureColorContrast) {
+    return <DonutChartWithContrast {...props} />;
+  }
+  return <DonutChartInner {...props} />;
+}
```
gamegee: [mid]: can't the condition be done without splitting the top-level components ?
Like just doing a memoized hook that compute the series

libs/ui-react-visualization/src/lib/Components/DonutChart/types.ts
```diff
@@ -125,6 +125,15 @@ export type DonutChartProps = {
...
   defaultActiveId?: string | null;
+  /**
+   * When `true`, each segment's `color` is nudged so its HSL lightness differs
+   * from the surface background by at least 20 units.
+   *
+   * Colors that already meet that minimum are left untouched. Useful when the
+   * segment palette includes colors that are close in lightness to the app background.
+   * @default false
+   */
+  ensureColorContrast?: boolean;
```
gamegee: ❤️

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -125,8 +130,45 @@ const LineChartContent = ({
...
+  const bgColor = useContrastBgColor();
+  return <LineChartInner {...props} bgColor={bgColor} />;
+}
+
+export function LineChart(props: Readonly<LineChartProps>) {
+  if (props.ensureColorContrast) {
+    return <LineChartWithContrast {...props} />;
+  }
+  return <LineChartInner {...props} />;
+}
```
gamegee: same question

libs/ui-react/src/lib/Components/core/AmountDisplay/AmountDisplay.test.tsx
```diff
@@ -23,7 +23,7 @@ describe('AmountDisplay', () => {
     );
     return Array.from(strips).map((el) => {
       const match = el.style.transform.match(/translateY\(-(\d+)%\)/);
-      return match ? parseInt(match[1]) / 10 : -1;
+      return match ? Number.parseInt(match[1]) / 10 : -1;
```
gamegee, line 26: 👍

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -17,16 +18,30 @@ import {
...
   accessibilityLabel = 'Donut chart',
   activeId: activeIdProp,
   defaultActiveId = null,
   onActiveIdChange,
   renderCenter,
   renderCenterActive,
-}: Readonly<DonutChartProps>) {
+  ensureColorContrast = false,
+  bgColor = '',
+}: DonutChartProps & { bgColor?: string }) {
```
gamegee: [low]: If you keep the same abstraction, maybe name could be a bit more explicit like `bgColorContrastReference`

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -12,18 +16,69 @@ import {
...
+    () =>
+      seriesProp.map((s) => ({
+        ...s,
+        ...(s.color &&
+          ensureColorContrast && {
+            color: getContrastSafeColor(s.color, bgColor, 35),
+          }),
+      })),
+    [seriesProp, ensureColorContrast, bgColor],
+  );
```
gamegee: Following this comment:
https://github.com/LedgerHQ/lumen/pull/823/changes#r3758076659

You could have your series computed like this:

```tsx
const resolvedSeries = useResolvedSeries({ series, ensureColorContrast })
```

Then the `useResolvedSeries` will be memoized ofc, and compute the right `bgColor` without the need to split-up components.

But maybe I am missing a performance flaw here ?

libs/design-core/src/lib/colorContrast/colorContrast.ts
aquelemiguel: sure, nice idea. i'll leave them unexported for the time being but i see the vision

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
aquelemiguel: ended up splitting because of this comment: https://github.com/LedgerHQ/lumen/pull/823#discussion_r3751124115

the issue this splitting is solving is that we don't need the subscription to theme if `ensureColorContrast` is false. we *could* pass `bgColor` from the outside but that's a bit leaky

i believe your proposed hook would still subscribe because of [rules of hooks](https://react.dev/reference/eslint-plugin-react-hooks/lints/rules-of-hooks), as they cannot be called conditionally

i dislike this splitting as well so i'm thinking: is what copilot reported such a performance issue? theme changes happen like once per session and it's likely super cheap. thinking it might be a non-issue. wdyt?

gamegee: Hmm I think `ThemeProvider` should be required for any component of Lumen.

To me its very standard approach when doing setup of a design-system, even if its not explicit in the docs of our components, it seems very expected setup to me.

aammami-ledger (changes requested):
I am not sure I understand the exemples in storybook, they look exactly the same to me 
(same for line chart btw)

Image

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
```diff
@@ -143,8 +185,22 @@ export function LineChart({
   magnetRadius,
   loading = false,
   emptyLabel = chartConfig.emptyState.defaultLabel,
+  ensureColorContrast = false,
```
aammami-ledger: maybe `enableColorContrast` would be a better prop name ?

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -23,9 +52,23 @@ export function DonutChart({
...
+  const bgColor = useContrastBgColor();
   const geometry = DONUT_GEOMETRY[size];
 
+  const series = useMemo(
+    () =>
+      seriesProp.map((s) => ({
+        ...s,
+        ...(s.color &&
+          ensureColorContrast && {
+            color: getContrastSafeColor(s.color, bgColor, 35),
```
aammami-ledger: Can we move the `35` to a const ? and why not use the default 20 ?

libs/ui-react-visualization/src/lib/Components/LineChart/LineChart.tsx
aquelemiguel: agreed, i was iffy about "ensure" as well

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
aquelemiguel: this was a mistake on my end, experimented with an override and left it
ended up dialling back to the default threshold of 20 for both donut and line charts

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChart.tsx
```diff
@@ -23,9 +27,27 @@ export function DonutChart({
   onActiveIdChange,
   renderCenter,
   renderCenterActive,
-}: Readonly<DonutChartProps>) {
+  enableColorContrast = false,
+}: DonutChartProps) {
+  const { colorScheme } = useTheme();
```
aammami-ledger: [high] We do not use the `useTheme` on react side it will throw if the app do not use `ThemeProvider` 🤔 

maybe do as we do in other places just use the `cssVar(...)`

gamegee: It requires an HEXA decimal, `cssVar()` is just an identity function and will return just the `--css-var` string

And the utility `getContrastSafeColor` requires an HEXA decimal string to compute properly the contrast

gamegee: Indeed, we are not using it on react-side that much, but to me it should be an implicit requirements, that `ThemeProvider` is required for the components to work

aammami-ledger (approved):

PR #825, opened 2026-08-07
refactor: restructure components folder

gamegee:


libs/ui-react/src/lib/Components/symbols/Icon/types.ts
aammami-ledger, line 1: Do we still expose the `IconProps` type  ? Cursor is pointing out that that's no longer the case

libs/ui-rnative/src/lib/Components/index.ts
```diff
@@ -1,49 +1,3 @@
...
-export * from './TextInput';
-export * from './ThemeProvider';
-export * from './Tile';
-export * from './TileButton';
-export * from './Tooltip';
-export * from './Trend';
-export * from './Utility';
+export * from './core';
+export * from './primitives';
+export * from './animations';
```
aammami-ledger, line 3: Same as above, maybe add `export * from './symbols/Icon';` ?

aammami-ledger (approved):

PR #826, opened 2026-08-10
fix(Button): default native buttons to type="button"

meriadec:
Native `<button>` elements default to `type="submit"`, so a Lumen button placed inside a `<form>` could submit it by accident. This defaults the native-button components to `type="button"`, matching what `TileButton`/`AvatarButton` already did.

- `BaseButton`, `MediaButton`, and `CardButton` now set `type="button"` before the props spread, so callers can still opt into `type="submit"`/`"reset"`.
- `asChild`/Slot renders are left untouched — no `type` is injected onto non-button elements.
- Web only; the React Native buttons use `Pressable`, which has no `type` concept.
- Docs updated (`Button` JSDoc + a "Form submission" section in `Button.mdx`), plus tests covering the default and the caller override.

### DIFF
--- a/libs/ui-react/src/lib/Components/Button/BaseButton.tsx
+++ b/libs/ui-react/src/lib/Components/Button/BaseButton.tsx
@@ -114,6 +114,11 @@ export const BaseButton = ({
       data-disabled={disabled || undefined}
       disabled={disabled}
       onClick={onClick}
+      // Default native buttons to type="button" so they do not accidentally
+      // submit a surrounding form. Placed before {...props} so a caller can
+      // still opt into type="submit"/"reset". Skipped for asChild, where the
+      // rendered element (e.g. an anchor) is not a native button.
+      {...(asChild ? {} : { type: 'button' as const })}
       {...props}
     >
       {loading && (
--- a/libs/ui-react/src/lib/Components/Button/BaseButton.tsx
+++ b/libs/ui-react/src/lib/Components/Button/BaseButton.tsx
@@ -114,10 +114,6 @@ export const BaseButton = ({
       data-disabled={disabled || undefined}
       disabled={disabled}
       onClick={onClick}
-      // Default native buttons to type="button" so they do not accidentally
-      // submit a surrounding form. Placed before {...props} so a caller can
-      // still opt into type="submit"/"reset". Skipped for asChild, where the
-      // rendered element (e.g. an anchor) is not a native button.
       {...(asChild ? {} : { type: 'button' as const })}
       {...props}
     >
--- a/libs/ui-react/src/lib/Components/Button/Button.mdx
+++ b/libs/ui-react/src/lib/Components/Button/Button.mdx
@@ -72,6 +72,17 @@ Buttons come in four different sizes:
 
 > **Important Note:** The `loading` and `disabled` states are **not coupled**. When `disabled` is set, all event handlers are deactivated. When only `loading` is set (without `disabled`), the button remains clickable. Combine both props if you need to prevent interactions during loading.
 
+### Form submission
+
+The Button defaults to `type="button"`, so placing it inside a `<form>` does **not** submit the form by accident. To use it as a submit or reset control, pass the native `type` explicitly:
+
+```tsx
+<form onSubmit={handleSubmit}>
+  <Button type='submit'>Save</Button>
+  <Button type='reset' appearance='gray'>Reset</Button>
+</form>
+```
+
 ## Responsive Layout
 
 Buttons dynamically adapt their width to the size of their content, while keeping the default paddings:
--- a/libs/ui-react/src/lib/Components/Button/Button.mdx
+++ b/libs/ui-react/src/lib/Components/Button/Button.mdx
@@ -74,7 +74,7 @@ Buttons come in four different sizes:
 
 ### Form submission
 
-The Button defaults to `type="button"`, so placing it inside a `<form>` does **not** submit the form by accident. To use it as a submit or reset control, pass the native `type` explicitly:
+On the native `<button>` render path, the Button defaults to `type="button"`, so placing it inside a `<form>` does **not** submit the form by accident. To use it as a submit or reset control, pass the native `type` explicitly:
 
 ```tsx
 <form onSubmit={handleSubmit}>
@@ -82,6 +82,8 @@ The Button defaults to `type="button"`, so placing it inside a `<form>` does **n
   <Button type='reset' appearance='gray'>Reset</Button>
 </form>
 ```
+
+When `asChild` is used, no `type` is injected, since the rendered element (e.g. an anchor) is not a native button.
 
 ## Responsive Layout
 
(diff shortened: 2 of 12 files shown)


libs/ui-react/src/lib/Components/Button/Button.test.tsx
```diff
@@ -58,4 +58,14 @@ describe('Button Component', () => {
...
+  it('should default to type="button" so it does not submit a form', () => {
+    render(<Button>Default type</Button>);
+    expect(screen.getByRole('button')).toHaveAttribute('type', 'button');
+  });
+
+  it('should let the caller override the type', () => {
+    render(<Button type='submit'>Submit</Button>);
+    expect(screen.getByRole('button')).toHaveAttribute('type', 'submit');
+  });
 });
```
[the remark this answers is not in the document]
meriadec: Good call. Added a Button `asChild` test asserting no `type` is injected onto Slot renders in `test(Button): cover asChild leaves type uninjected; clarify docs` (https://github.com/LedgerHQ/lumen/commit/c60c1460). (MediaButton has the same guard, but its nested `Slottable` breaks `React.Children.only` for a plain anchor child, so that asChild case is left out of scope here.)

libs/ui-react/src/lib/Components/Button/Button.tsx
```diff
@@ -21,6 +21,9 @@ const buttonVariants = cva('', {
  *
  * When in loading state, it displays a spinner. If an icon is provided without children, it renders as an icon-only button.
  *
+ * Defaults to `type="button"` so it does not accidentally submit a surrounding
+ * `<form>`. Pass `type="submit"` (or `type="reset"`) to opt into form behaviour.
+ *
```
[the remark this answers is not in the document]
meriadec: Clarified the JSDoc to scope the default to the native `<button>` render path and note that `asChild` renders get no injected `type`, in `test(Button): cover asChild leaves type uninjected; clarify docs` (https://github.com/LedgerHQ/lumen/commit/c60c1460).

libs/ui-react/src/lib/Components/Button/Button.mdx
```diff
@@ -72,6 +72,17 @@ Buttons come in four different sizes:
 
 > **Important Note:** The `loading` and `disabled` states are **not coupled**. When `disabled` is set, all event handlers are deactivated. When only `loading` is set (without `disabled`), the button remains clickable. Combine both props if you need to prevent interactions during loading.
 
+### Form submission
+
+The Button defaults to `type="button"`, so placing it inside a `<form>` does **not** submit the form by accident. To use it as a submit or reset control, pass the native `type` explicitly:
```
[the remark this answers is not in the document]
meriadec: Clarified the MDX to scope the default to the native `<button>` render path and added a note that `asChild` renders get no injected `type`, in `test(Button): cover asChild leaves type uninjected; clarify docs` (https://github.com/LedgerHQ/lumen/commit/c60c1460).

libs/ui-react/src/lib/Components/Button/BaseButton.tsx
```diff
@@ -114,6 +114,11 @@ export const BaseButton = ({
       data-disabled={disabled || undefined}
       disabled={disabled}
       onClick={onClick}
+      // Default native buttons to type="button" so they do not accidentally
+      // submit a surrounding form. Placed before {...props} so a caller can
+      // still opt into type="submit"/"reset". Skipped for asChild, where the
+      // rendered element (e.g. an anchor) is not a native button.
+      {...(asChild ? {} : { type: 'button' as const })}
```
gamegee: I would suggest to add that to the props types in `types.ts`, so that we can explicitly define the `@default` metadata and a small comment that will appear in storybook, but also in our future docs hub.

If you made that change, then I guess you can just remove that comment (already explained in jsdoc)

libs/ui-react/src/lib/Components/Button/Button.tsx
```diff
@@ -21,6 +21,11 @@ const buttonVariants = cva('', {
  *
  * When in loading state, it displays a spinner. If an icon is provided without children, it renders as an icon-only button.
  *
+ * On the native `<button>` render path it defaults to `type="button"` so it does
+ * not accidentally submit a surrounding `<form>`. Pass `type="submit"` (or
+ * `type="reset"`) to opt into form behaviour. When `asChild` is used, no `type`
+ * is injected, since the rendered element (e.g. an anchor) is not a button.
+ *
```
gamegee: same here, could be remove

libs/ui-react/src/lib/Components/MediaButton/MediaButton.tsx
```diff
@@ -151,6 +151,11 @@ export const MediaButton = ({
         className,
       )}
       disabled={disabled}
+      // Default native buttons to type="button" so they do not accidentally
+      // submit a surrounding form. Placed before {...props} so a caller can
+      // still opt into type="submit"/"reset". Skipped for asChild, where the
+      // rendered element (e.g. an anchor) is not a native button.
+      {...(asChild ? {} : { type: 'button' as const })}
```
gamegee: Same, I would just define it inside the `types` jsdoc

libs/ui-react/src/lib/Components/Button/BaseButton.tsx
meriadec: Done in `docs(Button): document type default in prop types, drop inline comments` (https://github.com/LedgerHQ/lumen/commit/3fb36f970bc773ebff1747969f8061edbcf3d17a) — `type` is now an explicit prop in `Button/types.ts` with `@default 'button'` JSDoc (so it shows in Storybook and the docs hub), and the inline comment was removed.

libs/ui-react/src/lib/Components/Button/Button.tsx
meriadec: Done in `docs(Button): document type default in prop types, drop inline comments` (https://github.com/LedgerHQ/lumen/commit/3fb36f970bc773ebff1747969f8061edbcf3d17a) — removed this JSDoc paragraph now that the default is documented on the `type` prop in `types.ts`. Kept the runnable `type="submit"` form example.

libs/ui-react/src/lib/Components/MediaButton/MediaButton.tsx
meriadec: Done in `docs(Button): document type default in prop types, drop inline comments` (https://github.com/LedgerHQ/lumen/commit/3fb36f970bc773ebff1747969f8061edbcf3d17a) — `type` is now an explicit prop in `MediaButton/types.ts` with `@default 'button'` JSDoc, and the inline comment was removed. Also applied the same to `CardButton` for consistency.

libs/ui-react/src/lib/Components/Button/Button.mdx
```diff
@@ -72,6 +72,19 @@ Buttons come in four different sizes:
 
 > **Important Note:** The `loading` and `disabled` states are **not coupled**. When `disabled` is set, all event handlers are deactivated. When only `loading` is set (without `disabled`), the button remains clickable. Combine both props if you need to prevent interactions during loading.
 
+### Form submission
```
gamegee, line 75: 🫶

gamegee (approved):

PR #827, opened 2026-08-10
feat(legend): introduce Legend component for visualization

aammami-ledger:
image

-----

image

### DIFF
--- a/apps/app-sandbox-rnative/src/app/(blocks)/Legends.tsx
+++ b/apps/app-sandbox-rnative/src/app/(blocks)/Legends.tsx
@@ -19,12 +19,12 @@ export default function Legends() {
         flexDirection: 'column',
         gap: 's24',
         width: 'full',
-        paddingLeft: 's16',
-        paddingRight: 's16',
+        paddingHorizontal: 's16',
       }}
     >
       <Base items={cryptoItems} />
       <DefaultColors />
+      <LongLabels items={cryptoItems} />
       <WithDonutChart segments={cryptoSegments} />
     </Box>
   );
@@ -78,11 +78,23 @@ const DefaultColors = () => (
   </Section>
 );
 
+const LongLabels = ({ items }: { items: LegendItem[] }) => (
+  <Section title='Long labels'>
+    <Legend
+      lx={{ maxWidth: 's176' }}
+      items={items.map((item, index) => ({
+        ...item,
+        label: index === 1 ? 'Wrapped Staked Ethereum Liquidity' : item.label,
+      }))}
+    />
+  </Section>
+);
+
 const WithDonutChart = ({ segments }: { segments: DonutSegment[] }) => (
   <Section title='With DonutChart'>
     <Box lx={{ flexDirection: 'row', alignItems: 'center', gap: 's24' }}>
       <DonutChart series={segments} />
-      <Legend items={segments} style={{ maxWidth: 176, flexShrink: 1 }} />
+      <Legend items={segments} lx={{ maxWidth: 's176', flexShrink: 1 }} />
     </Box>
   </Section>
 );
--- a/apps/app-sandbox-rnative/src/app/(blocks)/Legends.tsx
+++ b/apps/app-sandbox-rnative/src/app/(blocks)/Legends.tsx
@@ -62,14 +62,14 @@ const Section = ({
 
 const Base = ({ items }: { items: LegendItem[] }) => (
   <Section title='Base'>
-    <Legend items={items} />
+    <Legend series={items} />
   </Section>
 );
 
 const DefaultColors = () => (
   <Section title='Default colors'>
     <Legend
-      items={[
+      series={[
         { id: 'a', label: 'Segment A' },
         { id: 'b', label: 'Segment B' },
         { id: 'c', label: 'Segment C' },
@@ -82,7 +82,7 @@ const LongLabels = ({ items }: { items: LegendItem[] }) => (
   <Section title='Long labels'>
     <Legend
       lx={{ maxWidth: 's176' }}
-      items={items.map((item, index) => ({
+      series={items.map((item, index) => ({
         ...item,
         label: index === 1 ? 'Wrapped Staked Ethereum Liquidity' : item.label,
       }))}
@@ -94,7 +94,7 @@ const WithDonutChart = ({ segments }: { segments: DonutSegment[] }) => (
   <Section title='With DonutChart'>
     <Box lx={{ flexDirection: 'row', alignItems: 'center', gap: 's24' }}>
       <DonutChart series={segments} />
-      <Legend items={segments} lx={{ maxWidth: 's176', flexShrink: 1 }} />
+      <Legend series={segments} lx={{ maxWidth: 's176', flexShrink: 1 }} />
     </Box>
   </Section>
 );
--- a/apps/app-sandbox-rnative/src/app/(blocks)/Legends.tsx
+++ b/apps/app-sandbox-rnative/src/app/(blocks)/Legends.tsx
@@ -1,0 +1,88 @@
+import { Box, Text } from '@ledgerhq/lumen-ui-rnative';
+import { useTheme } from '@ledgerhq/lumen-ui-rnative/styles';
+import {
+  DonutChart,
+  Legend,
+  type DonutSegment,
+  type LegendItem,
+} from '@ledgerhq/lumen-ui-rnative-visualization';
+import type { ReactNode } from 'react';
+
+export default function Legends() {
+  const { theme } = useTheme();
+  const cryptoItems = getCryptoItems(theme.colors.crypto);
+  const cryptoSegments = getCryptoSegments(theme.colors.crypto);
+
+  return (
+    <Box
+      lx={{
+        flexDirection: 'column',
+        gap: 's24',
+        width: 'full',
+        paddingLeft: 's16',
+        paddingRight: 's16',
+      }}
+    >
+      <Base items={cryptoItems} />
+      <DefaultColors />
+      <WithDonutChart segments={cryptoSegments} />
+    </Box>
+  );
+}
+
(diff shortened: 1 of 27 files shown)


apps/app-sandbox-rnative/src/app/(blocks)/Legends.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const cryptoSegments = getCryptoSegments(theme.colors.crypto);
+
+  return (
+    <Box
+      lx={{
+        flexDirection: 'column',
+        gap: 's24',
+        width: 'full',
+        paddingLeft: 's16',
+        paddingRight: 's16',
```
aquelemiguel:
```suggestion
        paddingHorizontal: 's16',
```

libs/ui-rnative-visualization/src/lib/Components/Legend/types.ts
```diff
@@ -0,0 +1,34 @@
...
+   */
+  label?: string;
+  /**
+   * Swatch color.
+   * Falls back to the neutral default when omitted.
+   */
+  color?: string;
+};
+
+export type LegendProps = {
```
aquelemiguel, line 20: extend `StyledViewProps` for `lx` support?

libs/ui-react-visualization/src/lib/Components/Legend/LegendItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+  >
+    <span
+      aria-hidden='true'
+      data-testid='legend-swatch'
+      className='size-8 shrink-0 rounded-full'
+      style={{
+        backgroundColor: item.color ?? cssVar('var(--background-muted-strong)'),
+      }}
+    />
+    <span className='body-3 text-base'>{item.label ?? item.id}</span>
```
aquelemiguel: could this not be a bit too long so it overflows? should we add `min-w-0 truncate` here?

libs/ui-react-visualization/src/lib/Components/Legend/LegendItem.tsx
```diff
@@ -0,0 +1,26 @@
+import { cssVar } from '@ledgerhq/lumen-design-core';
+
+import type { LegendItem as LegendItemData } from './types';
+
+type LegendItemProps = {
+  item: LegendItemData;
+};
+
+// Internal, not exported from the barrel: the legend owns its own row markup.
+export const LegendItem = ({ item }: Readonly<LegendItemProps>) => (
```
aquelemiguel: react props are immutable by default and `Readonly` doesn't work recursively with nested objs, it's shallow. also, i'm not sure because of destructure into `item` the Readonly applies (as opposed to reading from like `props.item`) so i'd just remove this annotation

found the same pattern in other files

libs/ui-react-visualization/src/lib/Components/Legend/__stories__/Legend.mdx
```diff
@@ -0,0 +1,57 @@
...
+import { DocTable } from '../../../../../../ui-react/.storybook/components';
+import * as LegendStories from './Legend.stories';
+
+<Meta of={LegendStories} />
+
+# Legend
+
+A display-only key for chart series or segments. Each item shows a color swatch
+and a label. The legend is composable next to `DonutChart` and `LineChart` and
+does not drive or react to chart interaction state.
```
aquelemiguel: could be a bit clearer, took me a bit to understand.

"does not drive or react" reads like a feature when it should perhaps be an aside:

> For the time being, Legend is purely static: it does not react to nor change the DonutChart or LineChart's state and active segments.

something like that?

aquelemiguel: just read the "With DonutChart" section, i feel like that's way clearer:

> The legend is static for now and does not share activeId with the ring.

libs/ui-react-visualization/src/lib/Components/Legend/__stories__/Legend.stories.tsx
aquelemiguel, line 1: maybe truncation would also be interesting to showcase?

libs/ui-react-visualization/src/lib/Components/Legend/__stories__/Legend.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+> View in [Figma](https://www.figma.com/design/JxaLVMTWirCpU0rsbZ30k7/2.-Components-Library?node-id=16550-7583).
+
+## Usage
+
+Pass an array of `LegendItem`s. `DonutSegment[]` is assignable as-is; for
+`LineChart`, map each `Series` to `{ id, label, color: stroke }`.
```
aquelemiguel: obviously the no. of legend items should be the same as segments/series, but i could not find an internal check for this

if we're not checking it (should we not?) i'd vouch for a small note mentioning proper usage

libs/ui-react-visualization/src/lib/Components/Legend/__stories__/Legend.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+## Colors
+
+Without a `color` override, swatches fall back to the neutral default grey.
+
+<Canvas of={LegendStories.DefaultColors} />
+
+## With DonutChart
+
+Place the legend beside the chart. The legend is static for now and does not
```
aquelemiguel: a bit too barebones IMO, perhaps stress the fact consumers should layout legend themselves according to their usage, handle flex behaviours, and that our components do not do this for them

libs/ui-rnative-visualization/src/lib/Components/Legend/Legend.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (items.length === 0) {
+    return null;
+  }
+
+  return (
+    <View
+      accessibilityRole='list'
+      accessibilityLabel={accessibilityLabel}
+      testID='legend'
+      style={StyleSheet.flatten([styles.root, style])}
```
aquelemiguel: `style` prop already takes an array of styles, you shouldn't need to merge styles on every render

libs/ui-rnative-visualization/src/lib/Components/Legend/Legend.tsx
```diff
@@ -0,0 +1,43 @@
...
+    >
+      {items.map((item) => (
+        <LegendItem key={item.id} item={item} />
```
aquelemiguel, line 39: because the parent has accessibility role `list`, the children could have `listitem`, probably better screen reader support

libs/ui-rnative-visualization/src/lib/Components/Legend/LegendItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * Internal row: swatch + label. Not exported from the barrel — the legend owns
+ * its own row markup.
+ */
+export const LegendItem = ({ item }: Readonly<LegendItemProps>) => {
+  const styles = useLegendItemStyles(item.color);
+
+  return (
+    <View testID='legend-item' nativeID={item.id} style={styles.root}>
+      <View accessible={false} testID='legend-swatch' style={styles.swatch} />
+      <Text style={styles.label}>{item.label ?? item.id}</Text>
```
aquelemiguel: similarly to react, maybe `numberOfLines` makes sense here

libs/ui-rnative-visualization/src/lib/Components/Legend/LegendItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Internal row: swatch + label. Not exported from the barrel — the legend owns
+ * its own row markup.
+ */
+export const LegendItem = ({ item }: Readonly<LegendItemProps>) => {
+  const styles = useLegendItemStyles(item.color);
+
+  return (
+    <View testID='legend-item' nativeID={item.id} style={styles.root}>
```
aquelemiguel: is the point of this `nativeID` to forward it to `key` when inserting this into a list? i've only really seen this for harnessing in native UI tests. can this be removed?

aquelemiguel: we've got a hell of a lot of `testID` and if we could find a way to target these elements via the `by` APIs (`getByText`, `getByRole`, `getByLabelText`, etc) would reduce the bloat on our components

https://reactnative.dev/docs/testing-overview#testing-user-interactions

> Conversely, you should avoid:
> * making assertions on component props or state
> * testID queries

aquelemiguel (approved):
approved, excellent work abla 🥇

libs/ui-react-visualization/src/lib/Components/Legend/__stories__/Legend.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+ */
+export const WithDonutChart: Story = {
+  parameters: {
+    layout: 'centered',
+    backgrounds: { default: 'light' },
+  },
+  render: () => (
+    <div className='flex items-center gap-24'>
+      <DonutChart series={cryptoSegments} />
+      <Legend items={cryptoSegments} className='max-w-176' />
```
gamegee: [mid]: call it `series` too ? 
So we are aligned on the name, it fits with linechart, barchart & donutchart

gamegee (commented):
optional: I would suggest to add a story of LineChart with Legend (like in the bottom)
Could be nice

libs/ui-react-visualization/src/lib/Components/Legend/__stories__/Legend.stories.tsx
aammami-ledger: fair point, went with Items, because of the `LegendItems` internal subcomponent but yeah you are right.

PR #830, opened 2026-08-12
feat(BaseInput): add readOnly prop for AddressInput & TextInput

gamegee:
## Summary

- Promote `readOnly` to a documented, first-class prop on the React Native `BaseInput`, aligning the API with React's native `readOnly`. It prevents editing without applying the muted visual style of `disabled`, and hides the clear button while read-only.
- Add read-only stories for `AddressInput` and `TextInput` on both React and React Native, and document the read-only state in the RN MDX docs.
- Add a dedicated `AddressInputs` sandbox block (default, custom prefix, error, disabled, read-only) and register it; switch the `TextInputs` "Company" example to the new `readOnly` prop.
- Includes patch version plans for `@ledgerhq/lumen-ui-react` and `@ledgerhq/lumen-ui-rnative`.

## Test plan

- [ ] Storybook (react + rnative): AddressInput/TextInput show a `ReadOnly` story; read-only input is non-editable, keeps normal (non-muted) styling, and shows no clear button.
- [ ] RN MDX docs render the read-only Canvas/Source examples.
- [ ] Sandbox app: `AddressInput` block appears in the list; read-only variants behave as expected.
- [ ] `npx nx run-many --target=typecheck --all` passes.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/libs/ui-rnative/src/lib/Components/core/AddressInput/types.ts
+++ b/libs/ui-rnative/src/lib/Components/core/AddressInput/types.ts
@@ -1,7 +1,7 @@
+import type { ReactNode } from 'react';
 import type { BaseInputProps } from '../../internal/BaseInput';
-import type { BoxProps } from '../../primitives';
 
-export type AddressInputProps = Omit<BaseInputProps, 'prefix' | 'label'> & {
+export type AddressInputProps = {
   /**
    * Custom suffix element to show instead of the QR code icon.
    * Default suffix is a QR code scanner when empty (if onQrCodeClick provided), clear button when content.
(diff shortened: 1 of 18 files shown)


libs/ui-rnative/src/lib/Components/core/AddressInput/types.ts
```diff
@@ -14,11 +14,11 @@ export type AddressInputProps = Omit<BaseInputProps, 'prefix' | 'label'> & {
    * Custom prefix text to show instead of the "To:" prefix.
    * @default "To:"
    */
-  prefix?: string;
+  prefix?: ReactNode;
```
aammami-ledger, line 17: 👌

aammami-ledger (approved):
LGTM! ✅ 
Just added the chromatic tag to make sure we introduced no regressions 👌

PR #836, opened 2026-08-14
feat(Card): add appearance prop to CardFooter

aammami-ledger:
image


libs/ui-react/src/lib/Components/core/Card/Card.stories.tsx
```diff
@@ -226,6 +226,60 @@ export const StatesShowcase: Story = {
   ),
 };
 
+export const AppearanceShowcase: Story = {
```
gamegee: -> specific ? `FooterAppearance`

libs/ui-rnative/src/lib/Components/core/Card/types.ts
```diff
@@ -103,7 +103,16 @@ export type CardTrailingProps = {
   children: ReactNode;
 } & Omit<StyledViewProps, 'children'>;
 
+export type CardFooterAppearance = 'plain' | 'no-background';
+
 export type CardFooterProps = {
+  /**
+   * The visual appearance of the footer background.
+   * - `'plain'` (default): muted transparent background.
+   * - `'no-background'`: transparent, blends with the card surface.
```
gamegee: [mid]: Usually I don't like to duplicate all the values, we should get rid of that in the future.
AI and JSDoc parser are already able to output the typescript value

libs/ui-react/src/lib/Components/core/Card/types.ts
```diff
@@ -128,7 +128,16 @@ export type CardTrailingProps = {
   className?: string;
 } & ComponentPropsWithRef<'div'>;
 
+export type CardFooterAppearance = 'plain' | 'no-background';
+
 export type CardFooterProps = {
+  /**
+   * The visual appearance of the footer background.
+   * - `'plain'` (default): muted transparent background.
+   * - `'no-background'`: transparent, blends with the card surface.
```
gamegee: same

libs/ui-rnative/src/lib/Components/core/Card/types.ts
gamegee: Maybe add a ai rule for that ?

gamegee (approved):
✅ few coms, but good

libs/ui-rnative/src/lib/Components/core/Card/types.ts
aammami-ledger: okay

PR #842, opened 2026-08-18
feat(SegmentedControl): add arrow controls to scroll when overflowing

aquelemiguel:
Closes [DLS-953](https://ledgerhq.atlassian.net/browse/DLS-953).

## Demo
https://github.com/user-attachments/assets/8b2cf562-be1e-4b1d-ac18-d8b9132afd0d

[DLS-953]: https://ledgerhq.atlassian.net/browse/DLS-953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
+++ b/libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
@@ -1,8 +1,14 @@
 import { cn, useDisabledContext } from '@ledgerhq/lumen-utils-shared';
 import { cva } from 'class-variance-authority';
-import type { HTMLAttributes, ReactElement, ReactNode, RefObject } from 'react';
-import { useCallback, useLayoutEffect, useRef, useState } from 'react';
+import type {
+  ComponentType,
+  HTMLAttributes,
+  ReactElement,
+  ReactNode,
+} from 'react';
+import { useRef } from 'react';
 import { ChevronLeft, ChevronRight } from '../../symbols';
+import type { IconSize } from '../../symbols/Icon/types';
 import {
   SegmentedControlContextProvider,
   useSegmentedControlContext,
@@ -12,10 +18,12 @@ import type {
   SegmentedControlProps,
   SegmentedControlValue,
 } from './types';
+import type { PillLayout } from './usePillElementLayoutEffect';
 import {
   usePillElementLayoutEffect,
   useSegmentedControlSelectedIndex,
 } from './usePillElementLayoutEffect';
+import { useScrollArrows, useScrollSelectedIntoView } from './useScrollArrows';
 
 const segmentedControlStyles = {
   root: cva('relative flex flex-row items-center rounded-sm', {
@@ -65,7 +73,15 @@ const segmentedControlStyles = {
       ],
     },
   ),
-  arrowButton: cva(
+  scrollArea: cva('relative rounded-sm', {
+    variants: {
+      appearance: {
+        background: 'bg-surface',
+        'no-background': 'bg-transparent',
+      },
+    },
+  }),
+  arrow: cva(
     'absolute inset-y-0 z-20 flex cursor-pointer items-center justify-center rounded-sm py-8 text-muted transition-opacity duration-200 hover:text-muted-hover',
     {
       variants: {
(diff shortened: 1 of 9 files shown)


libs/ui-react/src/lib/Components/core/SegmentedControl/types.ts
```diff
@@ -25,11 +25,13 @@ export type SegmentedControlProps<
...
-   * - "fit": each tab sizes to its content
-   * - "fixed": all tabs share equal width, filling the container
+   * - `"fixed"`: all tabs share equal width, filling the container
+   * - `"fit"`: each tab sizes to its content (inline)
+   * - `"fit-controls"`: same as `"fit"` but adds left/right arrow buttons that
+   *   scroll the control when it overflows its container
    * @default 'fixed'
    */
-  tabLayout?: 'fit' | 'fixed';
+  tabLayout?: 'fixed' | 'fit' | 'fit-controls';
```
gamegee: [low/mid]: the name `fit-controls` feels not explicit to me, I was thinking of having `fit-responsive` or `fit-auto-scroll`, open to discussion

gamegee: On that specific API, I have a question:
- Should we let the consumers explicitely pass the `tabLayout: "fit-controls"` ?
- Or should we just enable the scroll and controls whenever the inner element width overlap its parent.

----
Follow me.
As part of the `SegmentedControl` we could have the following elements:
- An `outer div`, that will never exceed the width of its parent
- An `inner div`, that will contain all the `SegmentedControlItem` that can oveerlap the `outer-div`

Inside the component we could programmatically compute the difference of the width of both `outer` and `inner` div, e.g.

```tsx
const isControlsEnabled = tabLayout === 'fit' && innerDiv.width > outerDiv.width
// do stuff
```

gamegee: So it would mean by default, all the segmentedControl using `fit` variant will be responsive by default.

It could be a good approach, do you have opinions on that ? @aammami-ledger @aquelemiguel

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
```diff
@@ -147,30 +263,78 @@ export function SegmentedControl<
...
+            'relative rounded-sm',
+            appearance === 'background' ? 'bg-surface' : 'bg-transparent',
+            className,
+          )}
+        >
+          <div
+            ref={scrollRef}
+            className='scrollbar-none overflow-x-auto'
+            style={{
+              maskImage: `linear-gradient(to right, ${canScrollLeft ? 'transparent 0px, transparent 40px, black 72px' : 'black 0px'}, ${canScrollRight ? 'black calc(100% - 72px), transparent calc(100% - 40px), transparent 100%' : 'black 100%'})`,
```
gamegee: [mid]: You can split this part as a separate component, so it becomes easier to read:
- SegmentedControlMaskWrapper (name tbd)
- SegmentedControlArrowLeft
- SegmentedControlArrowRight


```tsx
<div>
  {showControls && 
    <SegmentedControlMaskWrapper>
      {chiildren}
      {pill}
  </SegmentedControlMaskWrapper>
</div>
```

```tsx
const SegmentedControlMaskWrapper = ({children}) => {
  return <div> // mask stuff
    {children}
    <SegmentedControlArrowLeft />
    <SegmentedControlArrowRight />
  </div>
}
```

So you can move down the `useLayoutEffect` in `SegmentedControlMaskWrapper` probably.
Splitting up the main component into smaller chunk would be easier to understand.

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
```diff
@@ -128,16 +206,54 @@ export function SegmentedControl<
...
+      return;
+    }
+    const buttons = container.querySelectorAll('button[role="radio"]');
+    const target = buttons[selectedIndex] as HTMLElement | undefined;
+
+    // scroll the selected item into view in case of "fit-controls" on mount
+    target?.scrollIntoView?.({ block: 'nearest', inline: 'nearest' });
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
gamegee: [low]: as it has been done for the pill, this could be split into a separate file.

Also the useLayoutEffect of line 228 could be instanciated directly in the useScrollArrows probably. Or maybe as a different hook `useScrollToSelectedItemOnMount`

aammami-ledger (commented):
When navigating with keyboard in the Fit with controls story the focus looks cropped compared to the other tabLayouts 

Image

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
```diff
@@ -128,16 +206,54 @@ export function SegmentedControl<
...
+    }
+    const container = scrollRef.current;
+    if (!container) {
+      return;
+    }
+    const buttons = container.querySelectorAll('button[role="radio"]');
+    const target = buttons[selectedIndex] as HTMLElement | undefined;
+
+    // scroll the selected item into view in case of "fit-controls" on mount
+    target?.scrollIntoView?.({ block: 'nearest', inline: 'nearest' });
```
aammami-ledger: [high] Wouldn't this scroll the page as well if the segmented control in in the bottom of the page ?

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
```diff
@@ -147,30 +263,78 @@ export function SegmentedControl<
...
+          <div
+            ref={scrollRef}
+            className='scrollbar-none overflow-x-auto'
+            style={{
+              maskImage: `linear-gradient(to right, ${canScrollLeft ? 'transparent 0px, transparent 40px, black 72px' : 'black 0px'}, ${canScrollRight ? 'black calc(100% - 72px), transparent calc(100% - 40px), transparent 100%' : 'black 100%'})`,
+              WebkitMaskImage: `linear-gradient(to right, ${canScrollLeft ? 'transparent 0px, transparent 40px, black 72px' : 'black 0px'}, ${canScrollRight ? 'black calc(100% - 72px), transparent calc(100% - 40px), transparent 100%' : 'black 100%'})`,
+            }}
+          >
+            <div
+              {...props}
```
aammami-ledger: [mid] all props goes in here while the `classname` is on the outer div,  this can be misleading to consumer, for instance setting  aWidth via `className` works, but via style={{ width }} does not.

Maybe we can move the classname here and add a new `outerClassname` ? cc @gamegee

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
```diff
@@ -147,30 +263,78 @@ export function SegmentedControl<
...
+              className={segmentedControlStyles.root({
+                appearance: 'no-background',
+                tabLayout: 'fit',
+              })}
+            >
+              {children}
+              {pill}
+            </div>
+          </div>
+          <button
```
aammami-ledger: [mid] Let's add `aria-hidden` when there are no arrows ? for both arrows, otherwise they will break a18y.

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
gamegee: I would always props drill **every** props to the top-level container 👍 

Except maybe for specific component such as inputs where some props are drilled to the inner input

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
aammami-ledger: https://github.com/user-attachments/assets/f285e2a7-25e9-45e2-a58b-31d6fa544335

libs/ui-react/src/lib/Components/core/SegmentedControl/types.ts
aquelemiguel: > [low/mid]: the name fit-controls feels not explicit to me, I was thinking of having fit-responsive or fit-auto-scroll, open to discussion

i'm open to perhaps `fit-scroll` as `fit-responsive` might not be explicit still and `fit-auto-scroll` could be a bit too long

aquelemiguel: > Or should we just enable the scroll and controls whenever the inner element width overlap its parent.

agreed it's more elegant to auto-detect and it's a very nice insight!

although i'm pondering if it's safer to push a conservative API now and merge the two tab layouts later on once we see real usage of the new controls. because we'd pass a new overflow wrapper on `fit` i'm afraid of somehow silently breaking existing usage.

maybe you both think it's alright to risk it, so i can go for it. LMK

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
aquelemiguel: ty, excellent suggestion, isolating the controls part
added in https://github.com/LedgerHQ/lumen/pull/842/commits/3389ae3e7849679cc06871c7b5b2882e6368ef46

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
aquelemiguel: moved the hook it as part of the splitting above. maybe the resulting component as part of your suggestion above might be too small to justify moving it to a diff file IMO

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
aquelemiguel: hum valid... perhaps `scrollIntoView` was the wrong choice here as i expected the scrolling to be contained internally. i have manipulated only `scrollLeft` as part of https://github.com/LedgerHQ/lumen/pull/842/commits/40b14d74b5dfa4a8c153160af4a2a03c077244d8, should not leak anymore

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
aquelemiguel: true and an oversight on my end, nice spot. clarifying:

* without controls, `radioGroup` is the top-level component and has `props` spread into it
* with controls, the top-level component is the ControlsWrapper and **should have `props` spread**

therefore, i conditionally remove the drilling from `radioGroup` and move it to the wrapper if `showControls` is true. added in https://github.com/LedgerHQ/lumen/pull/842/commits/a91b34d20dd5e84dc39d3c908c6ae018215bc26e

libs/ui-react/src/lib/Components/core/SegmentedControl/SegmentedControl.tsx
aquelemiguel: nice one, added in https://github.com/LedgerHQ/lumen/pull/842/commits/913ebb302dd8945ca6d83f05bcafb29fa0e507c9

aquelemiguel:
> When navigating with keyboard in the Fit with controls story the focus looks cropped compared to the other tabLayouts

@aammami-ledger excellent find abla, after a bit of trial and error had to add `focus-visible:-outline-offset-2`
pushed in https://github.com/LedgerHQ/lumen/pull/842/commits/5fb35bc24c7c62842b49ded6a4619c30ac2da5ee

libs/ui-react/src/lib/Components/core/SegmentedControl/useScrollArrows.ts
```diff
@@ -0,0 +1,104 @@
...
+    };
+  }, [update, scrollRef]);
+
+  return { canScrollLeft, canScrollRight };
+}
+
+/**
+ * Overflow state and the scroll action backing the left/right arrow buttons.
+ */
+export function useScrollArrows(scrollRef: RefObject<HTMLDivElement | null>): {
```
aammami-ledger, line 53: would be even better to add some test coverage for the hook no ?

gamegee: Its already tested on the component

aammami-ledger (approved):

PR #843, opened 2026-08-18
feat(DonutChart): Audit, performance check and minor fixes

aammami-ledger:
- pop active segments via SVG matrix so Android does not flicker a View transform before the ring redraws.
- Added `LongCenterLabel` component to handle long center titles and descriptions with truncation.
- Updated `DonutChartDescription` and `DonutChartTitle` to apply size-specific horizontal padding.
- Improved performance tests for DonutChart to ensure consistent rendering behavior.
- Enhanced styling for better responsiveness in both React and React Native versions.

### DIFF
--- a/libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
@@ -3,17 +3,17 @@ import { cva } from 'class-variance-authority';
 
 import { useDonutSizeContext } from './donutSizeContext';
 import type { DonutChartDescriptionProps } from './types';
-import { DONUT_GEOMETRY, getCenterMaxWidth } from './utils';
+import { getCenterContentInset, getCenterMaxWidth } from './utils';
 
 const descriptionVariants = cva('min-w-0 text-muted', {
   variants: {
     donutSize: {
-      md: 'px-4 body-3',
-      sm: '-mt-4 px-2 body-4',
+      md: 'body-3',
+      sm: '-mt-4 body-4',
     },
     layout: {
       text: 'truncate',
-      cluster: 'flex min-w-0 items-center gap-2',
+      cluster: 'flex items-center gap-2',
     },
   },
   defaultVariants: {
@@ -33,13 +33,17 @@ export const DonutChartDescription = ({
     consumerName: 'DonutChartDescription',
     contextRequired: false,
   });
-  const maxWidth = getCenterMaxWidth(DONUT_GEOMETRY[donutSize]);
+  const inset = getCenterContentInset(donutSize);
   const layout = isTextChildren(children) ? 'text' : 'cluster';
 
   return (
     <div
       ref={ref}
-      style={{ maxWidth, ...style }}
+      style={{
+        maxWidth: getCenterMaxWidth(donutSize),
+        paddingInline: inset,
+        ...style,
+      }}
       className={cn(descriptionVariants({ donutSize, layout }), className)}
       {...props}
     >
--- a/libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
+++ b/libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
@@ -1,19 +1,24 @@
-import { cn } from '@ledgerhq/lumen-utils-shared';
+import { cn, isTextChildren } from '@ledgerhq/lumen-utils-shared';
 import { cva } from 'class-variance-authority';
 
 import { useDonutSizeContext } from './donutSizeContext';
 import type { DonutChartDescriptionProps } from './types';
 import { DONUT_GEOMETRY, getCenterMaxWidth } from './utils';
 
-const descriptionVariants = cva('flex items-center gap-2 truncate text-muted', {
+const descriptionVariants = cva('min-w-0 text-muted', {
   variants: {
     donutSize: {
-      md: 'body-3',
-      sm: '-mt-4 body-4',
+      md: 'px-4 body-3',
+      sm: '-mt-4 px-2 body-4',
+    },
+    layout: {
+      text: 'truncate',
+      cluster: 'flex min-w-0 items-center gap-2',
     },
   },
   defaultVariants: {
     donutSize: 'md',
+    layout: 'text',
   },
 });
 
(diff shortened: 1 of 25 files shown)


libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartTitle.tsx
```diff
@@ -28,7 +41,7 @@ export const DonutChartTitle = ({
       ref={ref}
       style={{ maxWidth, ...style }}
       className={cn(
-        'truncate text-base',
+        titleVariants({ donutSize }),
         TITLE_TYPOGRAPHY[donutSize][size],
```
gamegee: [mid]: you can pass `className` to cva, it will concat it

Can `TITLE_TYPOGRAPHY` be computed as cva ?

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChartTitle.tsx
```diff
@@ -19,6 +19,11 @@ const TITLE_TYPOGRAPHY = {
   sm: { md: 'heading4SemiBold', sm: 'body2SemiBold' },
 } as const satisfies Record<DonutSize, Record<DonutTitleSize, Typography>>;
 
+const PADDING_X = {
+  md: 's4',
+  sm: 's2',
+} as const satisfies Record<DonutSize, 's2' | 's4'>;
```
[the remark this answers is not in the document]
gamegee: Hmm can be discussed, because both can evolve differently

Side-note: your styles constants are using `UPPER_CASE` and underscore syntax, is it common to our codebase ? 

Should we just `camelCase` ?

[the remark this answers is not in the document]
gamegee: [mid]: In rnative, I would suggest to use `useStyleSheet` if you need to combine variants and style, this is the common pattern used in most of components.

```tsx
const useStyles = ({ size }) => {
  return useStyleSheet((t) => {
    return ...
  }, [size])
}
```

```tsx
const resolvedStyles =  useStyles({ size })

<Text styles={[resolvedStyles, style]} {...props} />
```

It prevent mixing both `style` and `lx` composition in our implementations, so its easier to understand.

It also split the styles part from the rendering JSX part, which help with separation of concerns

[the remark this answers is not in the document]
gamegee: SideNode: `lx` is also slightly slower than `useStylesheet` in term of runtime performances

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
```diff
@@ -53,6 +70,7 @@ export function DonutChartDescription({
         flexDirection: 'row',
         alignItems: 'center',
         gap: 's2',
+        paddingHorizontal: paddingX,
```
gamegee: same comment, I would expect most rnative components to just have `useStylesheet` and no `lx`

libs/ui-rnative-visualization/src/lib/Components/DonutChart/DonutChart.performance.test.tsx
```diff
@@ -0,0 +1,218 @@
+import { beforeEach, describe, expect, it } from '@jest/globals';
```
gamegee, line 1: 🫶

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
```diff
@@ -29,12 +34,13 @@ export const DonutChartDescription = ({
     contextRequired: false,
   });
   const maxWidth = getCenterMaxWidth(DONUT_GEOMETRY[donutSize]);
+  const layout = isTextChildren(children) ? 'text' : 'cluster';
```
gamegee, line 37: 💯

libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.stories.tsx
gamegee: [low]: No more trailing icon, docs drift

gamegee: [low]: I would say, the comments for each stories may drift, I would be ok to totally nuke the comments on all the stories for linechart and donutchart, I don't think it adds value and its just more code that can drift and makes it very verbose.

- I do believe type interface + JSDoc + component comment is enough for devs
- And MDX already describe the stories for humans

wdyt ?

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartDescription.tsx
```diff
@@ -1,19 +1,24 @@
...
   variants: {
     donutSize: {
-      md: 'body-3',
-      sm: '-mt-4 body-4',
+      md: 'px-4 body-3',
+      sm: '-mt-4 px-2 body-4',
+    },
+    layout: {
+      text: 'truncate',
+      cluster: 'flex min-w-0 items-center gap-2',
```
gamegee: [low]: Double `min-w-0`

libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.stories.tsx
```diff
@@ -449,3 +488,138 @@ export const WithPreparedSeries: Story = {
...
+      )}
+    />
+  ),
+};
+
+/**
+ * Flips `loading` off so the reveal animation remounts the ring. Replay to
+ * watch CSS mask cost; the ring should remount once, not on every hover after.
+ */
+export const LoadingToData: Story = {
```
gamegee, line 509: [mid]: Not referenced in MDX

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartTitle.tsx
```diff
@@ -9,6 +10,18 @@ const TITLE_TYPOGRAPHY = {
   sm: { md: 'heading-4-semi-bold', sm: 'body-2-semi-bold' },
 } as const satisfies Record<DonutSize, Record<DonutTitleSize, string>>;
 
+const titleVariants = cva('min-w-0 truncate text-base', {
+  variants: {
+    donutSize: {
+      md: 'px-4',
+      sm: 'px-2',
+    },
```
gamegee, line 16: [mid]: Center inset now lives in two places per platform

-> The md=4px / sm=2px center inset is duplicated across four component files while both libs already own a single contentInset knob

libs/ui-react-visualization/src/lib/Components/DonutChart/__stories__/DonutChart.stories.tsx
aammami-ledger: Yes I will drop them all I agree +1
I will even add a rule to no longer add comments.

libs/ui-react-visualization/src/lib/Components/DonutChart/DonutChartTitle.tsx
aammami-ledger: Moved the values to config and added a small utils that serves the offset.

gamegee (approved):

PR #845, opened 2026-08-20
chore/dls-922-prod-peer-dev-deps

gamegee:


libs/ui-react/package.json
```diff
@@ -43,23 +43,23 @@
...
-    "react-i18next": "^14.0.0"
-  },
-  "peerDependencies": {
     "@base-ui/react": "^1.3.0",
+    "@ledgerhq/lumen-utils-shared": "0.1.11",
     "@radix-ui/react-checkbox": "^1.3.2",
     "@radix-ui/react-dialog": "^1.1.15",
     "@radix-ui/react-slot": "^1.2.3",
     "@radix-ui/react-switch": "^1.2.6",
     "@radix-ui/react-tooltip": "^1.2.8",
```
aammami-ledger, line 52: [mid] At the time we set this up, Radix was a peer so teams that already used it (ecommerce/LES) would share one version with Lumen, instead of Lumen nesting a second copy.

For cva, clsx, and tailwind-merge there was a second reason: we wanted consumers to install and use those utilities themselves.

That’s the model in [ADR 008 — Bundling Strategy](https://ledgerhq.atlassian.net/wiki/spaces/DLS/pages/6086361261/ADR+008+Bundling+Strategy).

If we want to change direction let's, please update/discuss the ADR before.

libs/ui-react-visualization/vite.config.ts
```diff
@@ -34,7 +34,20 @@ export default defineConfig(() => ({
...
+        'react/jsx-runtime',
+        '@ledgerhq/lumen-design-core',
+        '@ledgerhq/lumen-ui-react',
+        '@ledgerhq/lumen-utils-shared',
+        'class-variance-authority',
+        'clsx',
+        'tailwind-merge',
+        'd3-array',
+        'd3-scale',
+        'd3-shape',
```
aammami-ledger, line 49: ah cool, I have already added some of them in this hot-fix [PR](https://github.com/LedgerHQ/lumen/pull/848) 

I wanted to add them all including `d3` but it was way our of the scope of the fix, happy to see this 👌

libs/design-core/package.json
```diff
@@ -27,7 +27,9 @@
     "!dist/figma/*"
   ],
   "dependencies": {
-    "tailwindcss": "^4.1.17",
     "tslib": "^2.3.0"
+  },
+  "peerDependencies": {
+    "tailwindcss": "^4.1.17"
```
aammami-ledger, line 33: ok 👍

aammami-ledger (commented):
I likely won't be around when this gets merged, but I'd hope yo take some time to discuss this with @aquelemiguel and decide whether it makes more sense to keep clsx, tailwind-merge, and cva as peer dependencies — the original idea was to position them as standard tools to use alongside Lumen.

As for Radix, I think moving it to dependencies is fine, but keep in mind that some teams are using it internally and may be behind our current versions (which could result in two versions of the same package in their app).
But giving we dropping the usage anyways,  it's fine.

PR #847, opened 2026-08-21
feat(multiline-input): Add plan

aammami-ledger:


gamegee (commented):
As now the BaseInput will be either an `input` or a `textarea` in term of HTML elements, 

I would suggest to call it `BaseTextField`, it better convey the intent of the component, and that's also the choice Material did.

I guess at some point we should also go for:
- TextInput -> TextField
- AdressInput -> AdressField

But one thing after the other

gamegee (commented):
It seems the component use controlled & uncontrolled state internally
I would suggest to also use `useControllableState`

Part of the first small refactor / cleaning of the separation of concerns, we might also extract and improve the render part of the component, like line **134 -> 214** could be abstracted into smaller chunk.

Same for Rnative

docs/multiline-input.md
```diff
@@ -0,0 +1,249 @@
...
+
+Scrollbar appearance follows the same mapping as
+[Dialog.tsx](../libs/ui-react/src/lib/Components/core/Dialog/Dialog.tsx) line 242 —
+`scrollbarWidth === 'auto' ? 'scrollbar-custom' : 'scrollbar-none'`.
+
+#### 2. Autosize hook
+
+New `internal/BaseInput/useAutosizeTextarea/`, co-located like
+[useAutoWidthInput](../libs/ui-react/src/lib/Components/core/AmountInput/useAutoWidthInput/useAutoWidthInput.ts).
+Called only from `BaseInputMultiline`, so it needs no `multiline` guard:
```
gamegee, line 123: Discussed: Maybe do that same separation for Rnative, it could be easier to have a 1-1 matching, so the `useAutosize` is called only in specific condition

We could rename it `BaseInputMultilineAutosize`, like in material-ui, so it conveys the fact that height will be autosized based on:
- Count of line
- line-height / font-size passed to it

gamegee (commented):
For the plan, maybe the agent could take a look on how material-ui is optimizing the `TextAreaAutosize` - in term of computation + performance it could be interesting

PR #856, opened 2026-09-04
refactor(internals): move the Figma sync toolchain out of design-core

gamegee:
`libs/design-core` was a published package that also carried the entire Figma
sync toolchain, so tooling-only changes forced a package release, and the code
itself was never linted, typechecked or tested.

Introduce `internals/` for Nx projects that are local/dev-only and never
published, with `internals/sync-figma` as its first tenant: design tokens, SVG
symbols and the code-syntax write-back, all six targets plus the raw token JSON
that feeds them. `libs/design-core` now ships only production code; its
published `symbols/` export is unchanged.

The category follows the layout Coinbase's CDS uses for its own internal libs
(`@cds/figma-api` beside the published `@coinbase/cds-*`): a `private: true`
manifest scoped `@lumen/*`, with the Nx project name left unscoped
(`sync-figma`) so targets and the workflow read cleanly. Dev dependencies stay
in the root manifest — `dotenv` is added there, the one import that was
previously declared nowhere and resolved only via a hoisted transitive. "Never published" rests on three guards — `private: true` blocks npm,
`nx.json`'s `release.projects` glob (`libs/*`) excludes it from `nx release` and
`plan:check`, and `publish-dev-packages.yml` globs `libs/*/` too. Note CDS
enforces the import direction with nothing at all — no module-boundary rule, no
tags — relying on `private: true`, release-group exclusion and a depcheck
script, so a private lib landing in a published package's `dependencies` is
caught only by the consumer's install. The tag rule below is ours.

No `libs/*` project may import `internals/*`. That falls out of the existing
`scope:shared -> ['scope:shared']` allow-list plus the new `scope:internal`
tag, so **this PR changes `eslint.config.mjs` and `eslint.shared.mjs` not at
all** — both are byte-identical to `main`. Verified by planting an import in
`design-core`: relative paths and deep `@lumen/sync-figma/src/...` subpaths both
error. A bare `@lumen/sync-figma` import is not caught by lint, because the Nx
project name differs from the npm name, but fails to resolve at build.

Unpublished code keeps the dev lint profile: `internals/*` spreads
`sharedConfig`, same as `apps/*`, so its eslint config carries zero exceptions.
`prodConfig` bought only two rules here, one of which was illusory — it allowed
`scope:shared`, so it would have permitted the `utils-shared` import this branch
removes. A new drift-check invariant keeps the next internal project from
skipping the tag, the private manifest or the lint profile.

Bugs exposed by linting and typechecking this code for the first time:

- `designTokensEtl.ts` resolved prettier config from `process.cwd()`. Prettier
  searches upward from the argument's parent, so this only ever found the root
  config by accident of the old `cwd`; from the repo root it looked outside the
  repo and silently fell back to defaults.
- the same prettier action used `forEach(async …)`, returning before its writes
  landed and skipping files nondeterministically (4 formatted vs 7 once awaited).
- `figmaDownloadSvgs.ts` loaded `.env` from four levels above the repo root.
- one non-null assertion, and one implicitly-any index signature.

`toPascalCase` moved out of `@ledgerhq/lumen-utils-shared` into this project:
it existed only to serve the symbol codegen, so internals now depends on no
`libs/*` project at all. Version plans cover the three affected packages.

The four test files under the old `tools/` had never executed — the vitest
include covered `src/**` only. They now run: 20 tests passing.

Verified: the ETL and both symbol codegen targets reproduce their committed
output exactly, `design-core` builds with no tooling left in `dist/`, and the
boundary rule rejects a planted `internals` import from design-core and from
both apps.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


aammami-ledger (approved):
Really nice refactor 🫶 

Did you run the sync figma locally to make sure all is working as expected ?

Other from that LGTM!

PR #858, opened 2026-09-07
feat(MenuList): create MenuList & MenuListItem components

aquelemiguel:
Closes [DLS-968](https://ledgerhq.atlassian.net/browse/DLS-968).

[DLS-968]: https://ledgerhq.atlassian.net/browse/DLS-968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-rnative/src/lib/Components/core/MenuList/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  StyledViewProps,
+} from '../../../../styles';
+import type { IconSize } from '../../symbols/Icon/types';
+
+export type IconComponent = ComponentType<{
+  size?: IconSize;
+  color?: LumenTextStyle['color'];
+}>;
+
+export type MenuListItemAppearance = 'primary' | 'destructive';
```
zel-kass: [low]: I was wondering why the choice of props naming that differs from our usual `base` and `red`. I know we made an ADR about this semantics topic in the past but don't remember it and don't have access. I guess you followed the convention we had settled for in this case. Will read it again. Besides that, is the fact that no prop is documented with a JSDoc purposeful ?

libs/ui-rnative/src/lib/Components/core/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,146 @@
...
+          borderRadius: t.borderRadius.sm,
+          backgroundColor: t.colors.bg.baseTransparent,
+        },
+        pressed && { backgroundColor: t.colors.bg.baseTransparentPressed },
+      ]),
+      label: StyleSheet.flatten([
+        t.typographies.body2SemiBold,
+        {
+          flex: 1,
+          color: disabled
```
zel-kass: [low]: `resolveIconColor` and the `label.color` ternary have the same rule (disabled > destructive > base) over the same three tokens, one as token names and one as resolved theme values. so a new appearance has to be added in two places. Could we keep only the resolver, widen its name to something like resolveContentColor, type it NonNullable<LumenTextStyle['color']>, and have the label do t.colors.text[resolveContentColor(disabled, appearance)]? The icon call site stays as is.

libs/ui-rnative/src/lib/Components/core/MenuList/MenuList.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const { getByTestId } = render(
+      <TestWrapper>
+        <MenuListItem label='With Icon' icon={TestIcon} onPress={jest.fn()} />
+      </TestWrapper>,
+    );
+
+    expect(getByTestId('icon')).toBeTruthy();
+  });
+
+  it('renders icon with disabled appearance when disabled', () => {
```
zel-kass: [low]: both these tests assert the presence of the icon but not its colour, which can be misleading since they don't really test what they claim. You might want to assert the resolved colour just like what has been done for the text at line 84 instead.

libs/ui-rnative/src/lib/Components/core/MenuList/types.ts
```diff
@@ -0,0 +1,26 @@
+import type { ComponentType, ReactNode } from 'react';
+import type {
+  LumenTextStyle,
+  StyledPressableProps,
+  StyledViewProps,
+} from '../../../../styles';
+import type { IconSize } from '../../symbols/Icon/types';
```
zel-kass: [low]: Going past the barrel, could just import from `../../symbols/Icon`

zel-kass (approved):
Great work on this new baby. Pre approving as there is no real blocker, just some questions and quick fixes if they seem relevant to you 🦾

libs/ui-rnative/src/lib/Components/core/MenuList/types.ts
aquelemiguel: after discussion in daily let's do base | red, to keep it consistent with react version
let's raise the idea of "destructive" intent later

libs/ui-rnative/src/lib/Components/core/MenuList/types.ts
aquelemiguel: agreed, this was wrong on multiple other components.
pushed on https://github.com/LedgerHQ/lumen/pull/858/commits/fea8c045e3d21b67ddf4a6aab59295afcfd8aa89

libs/ui-rnative/src/lib/Components/core/MenuList/MenuList.test.tsx
aquelemiguel: good one

libs/ui-rnative/src/lib/Components/core/MenuList/MenuList.tsx
aquelemiguel: did exactly as you suggested, excellent one!

LaurineDevergne (approved):

PR #860, opened 2026-09-08
BREAKING_CHANGE(DotIcon, DotSymbol): apply getDotSize mapper pattern

aquelemiguel:
Closes [DLS-873](https://ledgerhq.atlassian.net/browse/DLS-873).

## Regarding the shims

Note the shims added in https://github.com/LedgerHQ/lumen/pull/860/commits/a8e19ae06d38ed32bb3c702bd7d00b49fd9113b4. `@ledgerhq/crypto-icons` imports `mediaImageDotSizeMap` and `mediaImageDotIconSizeMap` (see [usage in repo](https://github.com/search?q=repo%3ALedgerHQ%2Fcrypto-icons+mediaImageDotSizeMap&type=code)) which used by our Storybook build. Without those exports, Vercel won't build. Once the package drops the dependencies, we should remove the shims.

[DLS-873]: https://ledgerhq.atlassian.net/browse/DLS-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/ui-react/src/lib/Components/core/DotIcon/index.ts
+++ b/libs/ui-react/src/lib/Components/core/DotIcon/index.ts
@@ -2,9 +2,9 @@ export { DotIcon, dotIconSizeMap } from './DotIcon';
 export * from './getDotIconProps';
 export * from './types';
 
-// shim: @ledgerhq/crypto-icons imports this which used by our Storybook build...
-// remove once that package drops the dependency
-export const mediaImageDotSizeMap = {
+// shim: @ledgerhq/crypto-icons imports this
+// remove once the package drops the dependency
+export const mediaImageDotIconSizeMap = {
   40: 16,
   48: 20,
   56: 24,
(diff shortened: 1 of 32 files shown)


libs/ui-react/src/lib/Components/core/DotIcon/index.ts
```diff
@@ -1,7 +1,13 @@
...
+
+// shim: @ledgerhq/crypto-icons imports this
+// remove once the package drops the dependency
+export const mediaImageDotIconSizeMap = {
+  40: 16,
+  48: 20,
+  56: 24,
+  64: 24,
+  72: 32,
+} as const;
```
zel-kass, line 13: [low]: Previously was typed `as const satisfies Record<number, DotIconSize>`, is the change deliberate ? If not, maybe put it back while it is still in use. Same goes for DotSymbol index.ts.

zel-kass (approved):
Clean stuff. Already preparing the pr for crypto-icons lib so we can remove the shim ASAP. 😎

libs/ui-react/src/lib/Components/core/DotIcon/index.ts
aquelemiguel: fair!

LaurineDevergne (approved):

PR #862, opened 2026-09-11
fix(TextInput): fix potentially frozen label focus animation

aquelemiguel:
Closes [DLS-931](https://ledgerhq.atlassian.net/browse/DLS-931).

## Findings

* This appears to be a race condition issue with Lumen, not the consumer's code.
* There was a `useEffect` that started the label animation and returned an animation cancel as a cleanup, keyed on `isFloatingLabel` instead of on unmount.
* It's likely that on every focus change, React was sending a cancel and a new animation to the UI thread together, and the cancel won, thus the label would never move.
* It should resolve differently based on the scheduling load, so that probably explains why it consistently failed on LLM simulator but worked on physical and Lumen's sandbox.
* Therefore, **moved the cancel into its own unmount-only effect**.

## Patch on LLM sandbox

https://github.com/user-attachments/assets/c7249edf-54df-4032-a413-99a8e6287f17

## Regression on Lumen sandbox

https://github.com/user-attachments/assets/a046de80-57b5-4915-b00d-6a3959ddd139

[DLS-931]: https://ledgerhq.atlassian.net/browse/DLS-931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zel-kass (approved):
Gj, this one was pretty obscure ! 🫨

zel-kass (approved):
Gj, this one was pretty obscure ! 🫨

LaurineDevergne (approved):

PR #868, opened 2026-09-16
DLS-1027-chore(DotIcon, DotSymbol): remove mediaImage size-map shims

zel-kass:
## Description
[#860](https://github.com/LedgerHQ/lumen/pull/860) replaced the old size maps with `getDotIconProps` / `getDotSymbolProps`, but kept `mediaImageDotIconSizeMap` and `mediaImageDotSizeMap` as temporary re-exports so `@ledgerhq/crypto-icons` (and therefore Storybook / Vercel) would still build.
That consumer now uses the resolvers ([LedgerHQ/crypto-icons#143](https://github.com/LedgerHQ/crypto-icons/pull/143), published as `2.0.6`), so the shims can go. This drops them from the web and RN barrels, removes the tests that only existed to cover those exports, and bumps the workspace `@ledgerhq/crypto-icons` pin from `^2.0.4` to `^2.0.6` so Storybook does not import the deleted maps.
Callers should keep using:
```tsx
<DotIcon {...getDotIconProps('mediaImage', parentSize)} />
<DotSymbol {...getDotSymbolProps('mediaImage', parentSize)} />
```


.nx/version-plans/version-plan-1789486926126-ui-react.md
aquelemiguel, line 1: i don't think this one needs to be tagged as a breaking change (nor rnative) IMO given the one before already was!
just a simple chore version plan should suffice, no?

aquelemiguel (approved):

.nx/version-plans/version-plan-1789486926126-ui-react.md
zel-kass: Absolutely !

PR #870, opened 2026-09-16
DLS-1032-fix(Icons): Icons stroke width scaling

zel-kass:
## Context

Icons are exported from Figma on a 16×16 artboard, so every icon ships `viewBox="0 0 16 16"`. SVG `stroke-width` is expressed in **viewBox user units**, which the renderer multiplies by `size / 16`. Our design tokens (`--icon-border-width-*`) are the desired **painted pixel** weights, and they grow sublinearly with size for optical balance.

## Issue

Two stacked bugs meant painted stroke widths never matched Figma above 16px.

**1. The generator baked `stroke-width="1.3"` onto every `<path>`.** That's the correct value for a 16px artboard, but on the path it scales linearly with size and overrides anything the `Icon` wrapper sets.

| Size | Token (expected) | Painted | Error |
| --- | --- | --- | --- |
| 16 | 1.3 | 1.30 | exact |
| 24 | 1.8 | 1.95 | +8% |
| 40 | 2.2 | 3.25 | +48% |
| 56 | 2.8 | 4.55 | +63% |

**2. Applying the token directly on the `<svg>` doesn't work either.** The token is a pixel value, but `stroke-width` is interpreted in user units, so it gets multiplied by `size / 16` too — a 2.8 token painted as 9.8px at size 56.

## Fix

Strip the hardcoded attribute, then pre-divide the token so the renderer's multiplication cancels out:

`declared = token ÷ (size / 16)`  →  `painted = token`

- **Generator** (`generate-symbols.mts`): SVGO `removeAttrs` strips `stroke-width` from paths so they inherit from the `<svg>` wrapper.
- **Web** (`createIconUtilities.ts`): `.icon-stroke-24` now emits `calc(var(--icon-border-width-24) / 1.5)`. The division happens in CSS so the custom property stays intact and brand overrides still flow through. `ui-react/Icon.tsx` is unchanged — it still just applies `icon-stroke-{size}`.
- **React Native** (`ui-rnative/Icon.tsx`): one expression, `theme.icon.borderWidth[sizeKey] * (16 / size)`.

All 8 sizes now paint exactly their token on both platforms.

## Known trade-off

3 of 296 icons (`CurveUp`, `CurveDown`, `Ar`) ship `viewBox="0 0 17 16"`. Rendered into a square box their true scale is `size / 17`, so they land ~6% thin (1.69px vs 1.80px at size 24) — well under a device pixel. The alternative was parsing each icon's `viewBox` string at render time plus a public helper in `utils-shared`; not worth it. The clean long-term fix is normalising those three artboards in Figma.

## Review notes

The ~500 changed icon files are pure generator output (one deleted `stroke-width` line each). The actual change is 3 files: `generate-symbols.mts`, `createIconUtilities.ts`, `ui-rnative/Icon.tsx`.


aquelemiguel (changes requested):
change makes sense in isolation but as discussed, let's block this until designers realise how all components are affected

aquelemiguel (approved):

PR #871, opened 2026-09-18
DLS-1023-feat(Toast): Introduce Toast component

zel-kass:
Adds a Toast system for React so apps can confirm actions and surface status without blocking the UI.

Mount ToastProvider once and trigger toasts with useToast. Queue, timer, and controller logic live in utils-shared so React Native can reuse the same API later.

- Appearances: info, success, warning, error, plus a loading spinner
- Default timing: info/success auto-dismiss after 5s; warning, error, and loading persist until dismissed or updated
- Imperative API: notify / appearance shortcuts, update (same id), dismiss / dismissAll, and promise for loading → success/error
- Optional single action (does not dismiss) and optional close (dismissible)
- FIFO queue: up to maxItems visible (default 3), configurable position, hover/focus pauses timers
- Accessibility: polite vs assertive live regions by status; focus stays in the app
- Also includes Storybook docs, Code Connect, i18n for the close label, and a toast z-index token (150).


libs/ui-react/src/lib/Components/core/Toast/ToastProvider.tsx
```diff
@@ -0,0 +1,188 @@
+import {
```
gamegee: ## PR feedback: don't nest ToastProvider inside ThemeProvider

**A few reasons to decouple instead:**

Can't fire from outside the tree. Redux middleware, thunks, RTK Query callbacks, sagas — none of that code is "inside" a context provider, so useToast() is unreachable there. Any consuming app doing async side effects outside components loses toast entirely.

The escape hatch is a known trap. Chakra ships this exact shape and had to add createStandaloneToast() for out-of-tree calls — which then hit a real bug (chakra-ui#8800) where the standalone root can't see outer app context. Nesting works until you need the exit, and the exit reintroduces the problem.

No cross-platform upside. RN/web still need platform-specific portal logic regardless of tree position, and coupling means every consumer must take theme + toast together, with no way to use one without the other.

Suggested fix: module-level singleton (Observer/pub-sub), same pattern as Sonner, react-hot-toast, and Uber Base Web's toaster util — no context, callable from anywhere:

```tsx
class ToastObserver {
  private subs: Array<(t: Toast) => void> = [];
  subscribe(fn: (t: Toast) => void) {
    this.subs.push(fn);
    return () => { this.subs = this.subs.filter(s => s !== fn); };
  }
  publish(t: Toast) { this.subs.forEach(fn => fn(t)); }
}
export const toastObserver = new ToastObserver();
export const toast = {
  success: (msg: string) => toastObserver.publish({ type: 'success', msg }),
  error: (msg: string) => toastObserver.publish({ type: 'error', msg }),
};

// JSX part
<ThemeProvider theme={theme}>
  <App />
  <Toaster />  {/* sibling, subscribes on mount, config as own props */}
</ThemeProvider>


// Usage in features in react components or in redux separate files
toast.success()
```

libs/ui-react/src/lib/Components/core/Toast/Toast.mdx
```diff
@@ -0,0 +1,148 @@
...
+import { ToastProvider } from '@ledgerhq/lumen-ui-react';
+
+function App() {
+  return (
+    <ToastProvider position='bottom-right' maxItems={3}>
+      <Routes />
+    </ToastProvider>
+  );
+}
+```
```
gamegee: [low]: Maybe this should come right after the controls section at the top, since its the mandatory setup.
wdyt ?

PR #875, opened 2026-09-22
feat(MenuList): create MenuListSwitchItem subcomponent

aquelemiguel:
Closes [DLS-1028](https://ledgerhq.atlassian.net/browse/DLS-1028).

[DLS-1028]: https://ledgerhq.atlassian.net/browse/DLS-1028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ui-rnative/src/lib/Components/core/MenuList/types.ts
```diff
@@ -47,3 +47,34 @@ export type MenuListItemProps = {
...
+  /**
+   * Icon rendered before the label.
+   */
+  icon?: IconComponent;
+  /**
+   * Prevents interaction and dims the content.
+   *
+   * @default false
+   */
+  disabled?: boolean;
```
gamegee: [mid]: For these 3 props, I think we can just create a `MenuItemBaseProps` that is extended in your type, it prevent duplication of types & JSDoc.

Another solution is to directly extend using a `Pick<T>`

libs/ui-react/src/lib/Components/core/Menu/types.ts
```diff
@@ -188,6 +188,41 @@ export type MenuCheckboxItemProps = {
...
+   */
+  onCheckedChange?: (checked: boolean) => void;
+
+  /**
+   * Whether the menu closes when this item is clicked.
+   *
+   * @default false
+   */
+  closeOnClick?: boolean;
+} & Omit<MenuItemProps, 'appearance' | 'closeOnClick'>;
```
gamegee: 👍 like this

[low]: maybe a Pick would be more precise, as you will don't end up extending incorrect props if someone add new props.

More resistant to future updates

libs/ui-rnative/src/lib/Components/core/MenuList/MenuList.tsx
```diff
@@ -160,3 +163,91 @@ const MenuListItemInner = ({
...
+  label,
+  icon: Icon,
+  disabled: disabledProp = false,
+  checked: checkedProp,
+  defaultChecked = false,
+  onCheckedChange: onCheckedChangeProp,
+  accessibilityState,
+  lx,
+  style,
+  ref,
```
gamegee: [low]: these 3 props could be removed, then these will be drilled as part of the `...props`

gamegee (commented):
Docs improvements ideas:
1. Maybe have a full showcase in the BottomSheet would make sense too ?

2. Since on the `Menu` for react, we have multiple item types for selected state (radio, checkbox, switch); We could have a story showcase that show the item without menu (pure UI) instead of multiple story:
- CheckboxItem
- RadioItem
- SwitchItem
Then add the SwitchItem in the complete showcase story of Menu 

---
No blockers, just minor docs improvement, WDYT ?

